PR #25, opened 2023-11-10
Update test_serialize.py

ehthiede:
Updated the API in test_serialize from "compute" to "evaluate": I think this is just a holdover from an old API?


mjo22 (commented):
Yes, that's right! I missed this when moving over the API.

mjo22 (approved):

PR #35, opened 2023-12-12
Big overhaul for the API. It's not perfect, but a lot of awkwardness in the design should be much better now.

mjo22:
This removes the
- The ``PipelineState`` container. This was pretty confusing in my opinion
- The ``ScatteringImage``, ``OpticsImage``, and ``DetectorImage`` model. This was pretty awkward I think. The purpose of these was to compute images at different stages of the image formation pipeline, but this can just be done with null models, like ``NullOptics``.
- The awkward ``scatter`` methods I had lying around everywhere. Now, only ``ScatteringConfig`` has a ``scatter`` method. This functionality is taken over by ``ImagePipeline.render``

This adds
- An ``Instrument`` object. This holds the ``Optics``, ``Exposure``, and ``Detector``.
- The ``Pose`` to the ``Specimen``. This makes a lot of sense I think, and it turns out to make some code clean-up very very nice!
- The ``Ice`` model to the ``ImagePipeline``. This did not belong in the ``Instrument`` and it did not belong in the ``Specimen``. Therefore, this was the only place! This is called ``ImagePipeline.solvent``.

Some things aren't great. But, it's much better I think!


mjo22:
Adding an ``Assembly`` base class if users want to use the functionality that I wrote for ``Helix``.  Also, better public vs. private practices, better typing conventions, and some other miscellaneous reorganization and documentation fixes.

mjo22:
Closing, merge #92 instead

PR #50, opened 2023-12-15
implement dataloader for starfiles

DSilva27:


mjo22:
Deleting the api-update branch as this is long outdated. We can make an updated version of this pull request

PR #63, opened 2023-12-18
Overhaul the cryojax API, improve packaging

mjo22:


mjo22:
This will also merge prototypes for a helical simulator and image formation directly from atomic models. These are in active development and need unit testing

PR #90, opened 2023-12-23
Simulate xyz

ehthiede:
Added functionality for loading pdb's into voxel densities.


ehthiede:
The more I'm looking at this I'm a) not happy with my naming conventions, and b) worried that I'm actually conflicting with conventions in the main code.  Will close this and re-open with an improved version of the code.

Issue #91, opened 2023-12-24
Coordinate nomenclature is unclear and needs refactoring


ehthiede:
Currently, many things are called "coordinates."  

- in VoxelGrid `coordinates` is a plane of frequencies in _Fourier_ space
- in VoxelCloud  `coordinates` (I think?) are an N x 3  of points in _real_ space.  (This matches the use in AtomCloud, but technically there they are different objects, as 
- `coordinates.py` the routine `make_coordinates` returns an (N1 x N2 x N3 x 3) Array giving the real-space positions  (It calls _fftreqs, which also doesn't return fft frequencies if `Real=True`, which is a separate issue.)

I think it would be good to give these different things distinct, descriptive names so it is clear what each one is.

**Possible Solution**
It seems to me that we have 3 formats for storing this sort of data, each of which can be in either Real or Fourier space:
1. A grid of  (N1 x N2 x N3 x 3).  This gives the (x,y,z,) or (k_x, k_y, k_z) value at each grid point in regular 3D grid.
2. A grid of (N1 x N2 x 1 x 3).  This gives the (x,y,z,) or (k_x, k_y, k_z) value at each grid point in a regular 2D grid located on a plane in a 3D space.
3. An array of size (N x 3).  This gives the (x,y,z,) or (k_x, k_y, k_z) value at a bunch of points not located on a regular grid.

We can come up with a name for each, and then prepend "coordinate_" or "frequency_" to make it clear if we are in real or Fourier space.  One option would be.

1. System 
2. Plane
3. List

In this case, a set of all the points in a 3d volume in real space would be a "coordinate_system", in Fourier space it would be a "frequency_system", for the 2D system we would have a "coordinate_plane" or "frequency_plane", etc.

Note: I'm purposefully avoiding "Volume" for number 1 to avoid confusion with the use of the term to refer to the reconstructed electron density.  Also, I'm not sold on having these are the best three words, but they are there to get the juices flowing.

mjo22:
This is a good point. Note that for 2D coordinates, I do make the distinction between real and fourier space: https://github.com/mjo22/cryojax/blob/main/src/cryojax/simulator/manager.py

[...]

PR #92, opened 2023-12-25
Simulate xyz

ehthiede:
Second draft of the atomic structure to density code, hopefully the conventions are compatible now.  Loading into VoxelGrids should now be supported as well, as should loading from .cif files.

Some notes:

- There are a few unnecessary calls where I regenerate coordinate systems that already exist: eventually this will be refactored out, but for now I prioritized functioning over clean.
- Naming conventions should be revisited pending conclusion of #91 .
- I did have to make a minor change in the fft code since .dim doesn't work for jax's compiled ArrayImpl object.  Hopefully this didn't break anything elsewhere.


mjo22 (approved):
This looks fantastic. 

On the .dim issue: this was a bug on my part. This should have said ft.ndim and ift.ndim—your `dim = len(ft.shape)` does the same. I actually don’t use the rfftn and irfftn routines in the code anywhere yet, I just added these recently because of plans to eventually make things more memory-conservative.

One question this is making me have—what do you think about `Voxels.from_file`? This routine could be kept if we had some logic that chose a loader based on the file extension. Another route would be to just get rid of it all together.

Also, what do you think about the abstract methods and dataclass fields in the `Voxels` base class? For one, #91 means removing the “coordinates” AbstractVar. The only benefit of having these AbstractVars is if external objects should expect something like “weights” to be a field of something where isinstance(density, Voxels) is True. Right now, we could go either way, but in general these AbstractVars are a useful pattern for type checkers and developers.

For another, `Voxels.from_mrc` is an abstract method but other loaders are not. Should other loaders be added as abstract methods, or does it make sense as is?

We can definitely merge. We should resolve #91 before merging into main.

mjo22:
Another comment: it’s a little weird that I create instantiation dicts in load_voxels.py. I like that the new constructors don’t do this. In general, the load_voxels.py and from_mrc constructors should be rethought a little I think. We should update these as well?

ehthiede:
Thanks for the comments! I'm going to try and refactor the code to go with `coordinate_grid/plane/list` and then merge into `simulate_xyz`.  We can change the nomenclature to something else later, but having my names be more descriptive than they are right now should help with later refactorings.

Would you be ok with merging to main after that?  I'm a little worried about our branches drifting too far apart.  Alternatively, if you want to protect main a little more, we could start a develop branch: your call.

mjo22:
Yes this sounds good and let’s merge into main after this. Also, let’s just worry about the 3D coordinates for now—this shouldn’t break much in the code. For this, each of the VoxelGrid, VoxelCloud, FourierSliceExtract, and NufftProject will need to be modified. I don’t think anything else. I can go through renaming the 2D coordinate systems later, which will break more in the code (unless you would like to do so!).

Issue #88, opened 2023-12-23
It is confusing to have two pixel sizes


mjo22:
It is confusing to have two different pixel sizes in the model. Currently, there is the `ScatteringModel.pixel_size` and the `Detector.pixel_size`. The detector pixel size is used to interpolate to a new pixel size that is the correct one at the detector. However, this is really only something you would want to do with a voxel-based electron density. If imaging directly from atoms, it would make sense to rasterize at a different pixel size.

The issue runs a little deeper. As mentioned when imaging directly from atoms, the `ScatteringModel` pixel size is a parameter. When imaging from voxels, the `ScatteringModel` pixel size must be correctly configured from the electron density voxel size. Then, the “parameter” aspect of the pixel size is offloaded to the `Detector` pixel size. There should be one place where the pixel size is treated as a parameter. Therefore, I propose

- Get rid of `Detector.pixel_size`.
- Move pixel size interpolation to the `ScatteringModel`. Then, write some logic in the base class that understands if the density representation is voxel-based, interpolate to the new pixel size. Then, somehow we must also turn off the interpolation when we just want pixel size equal to voxel size.

On one hand though, it is intuitive that the pixel size is a parameter of the detector. This is nice. But by the time an image is computed by the `ScatteringModel`, the pipeline should no longer care whether that image was computed from a voxel-based or atom-based electron density representation.

mjo22:
Interesting idea: the pixel size logic mentioned above might be a lot easier to write if the `ScatteringModel` is given the `ElectronDensity` upon instantiation. The `pixel_size` can then be a property, where it is taken to be the `voxel_size` if some other parameter is set to `None`.

Then, the `Specimen` gets the `ScatteringModel` and so on. This would make a lot of sense I think actually.

mjo22:
Also, the pixel size interpolation methods would make most sense in the `ImageManager`.

Issue #51, opened 2023-12-16
Specimen.resolution refactoring


mjo22:
The name resolution is confusing because it sounds like the resolution of a reconstruction.

My leading candidate for renaming is Specimen.discretization, but it doesn’t feel perfect. Whatever it’s changed to should intuitively make sense as a property of a biological specimen (or an electron density, if we move it here).

At its core, this variable is the scale that we want to look at our image in “free space”, e.g. imagining that we had the ability to look at it with no optics, detector, or microscope.

More concretely, this is meant to represent the discretization of a 2D image. If we want to have separate a parameter for that of a voxel grid, maybe this should be another parameter explicitly (then we can do things like downsampling). We should just make sure the user doesn’t have to manually configure too many scales (see how the Detector treats the pixel size as optional for this reason).

If we want to move this parameter to the ElectronDensity base class, we will need to look at the ImagePipeline and ScatteringConfig.scatter to see what refactoring needs to be done

mjo22:
This can potentially be closed now with #77, #78 #79, and #81. pixel_size is now a parameter of the ``ScatteringModel`` and voxel_size is now a parameter of the ``ElectronDensity``. Additionally, the ``Detector`` gets a pixel_size that can be optionally set. The reason for this additional pixel size is 1) Detectors have pixel sizes and 2) We will want to do gradient descent on very small changes in pixel size in experimental data. Pixel sizes from microscope alignments can be slightly off.

mjo22:
In the future, we could now rasterize at a voxel_size, downsample to the ScatteringModel pixel_size, and then interpolate in small changes in the Detector pixel_size. Or, we can rasterize directly at the ScatteringModel pixel_size.

mjo22:
@ehthiede these changes should help with #29

mjo22:
Also see here #88. I think we are close to having this make sense in a unified way for atom and voxel based imaging. But, this confusion needs to be cleared up.

mjo22:
Proposing that this can be closed with #93

PR #93, opened 2023-12-25
Pixel size refactor

mjo22:
Addressing #88 and #51. Now, `pixel_size` is a parameter in the `ScatteringModel`. For voxel-based densities, this is an independent parameter from the voxel size that interpolates the voxel size to a new pixel size only if the two values are different. This functionality used to be in the `Detector`, but it was way too confusing and hacky to have it here. For future atom-based densities, this `pixel_size` parameter can be used directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,18 +37,19 @@ The [jax-finufft](https://github.com/dfm/jax-finufft) package is an optional dep
 
 Please note that this library is currently experimental and the API is subject to change! The following is a basic workflow to generate an image with a gaussian white noise model.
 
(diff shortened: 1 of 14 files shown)


ehthiede (approved):
I like the changes, in particular the way the electron stack is handled and the new use of ClassVars!

mjo22:
Noting here that there may be a bottleneck using the `__getitem__` routine of the `ElectronDensity`, even after jitting… but the code is cleaner than before so now I am punting this issue for a later time.

Issue #52, opened 2023-12-16
Rename GaussianImage


mjo22:
It may make sense to rename this class to something more intuitive, since its main purpose is to compute a likelihood. 

In doing this, would it make sense to make a base class for a probabilistic model? Should we make an abstract class that subclasses ImagePipeline, adding a log_probability method?

mjo22:
Created `Distribution` object with #64. Strictly speaking, `GaussianImage` is an independently but not identically distributed gaussian for arbitrary detector and noise variance functions. It could be re-named something along the lines of `IndependentFourierGaussian`…?

mjo22:
Closing with #100

Issue #68, opened 2023-12-19
Clean up superfluous FFTs in the pipeline


mjo22:
Right now I am a little loose with how many times I transform 2D images back and forth between real and fourier space. Can this be made more efficient? Also, can we make it so we only use the modes on the half plane for 2D and 3D FFTs of real-valued functions (i.e. with `jax.numpy.rfftn` and `jax.numpy.irfftn`)?

mjo22:
This will involve looking at the `extract_slice` function in the fourier slice theorem code, and the `GaussianImage.variance` function. I have ad-hoc things in there to make sure I don’t do redundant computation, but this should just be taken care of when computing wave vectors and FFTs.

Also, the `cryojax.utils.fft` module should instead have conventions for `rfftn` and `irfftn` using `jnp.fft.rfftn` and `jnp.fft.irfftn`.

ehthiede:
Relatedly, we could probably also clean up make_coordinates so it just builds the real-space coords using linspace: this avoids a slightly unclear move to Fourier space and back. and streamlines the code a bit.

mjo22:
Right now, I just use a jnp.fft.fftshift(jnp.fft.fftfreqs(…)) in the _fftfreq1D function instead of a jnp.linspace to be extra sure that real-space coordinate systems match exactly what is expected of an inverse FFT. I may not understand, what do you mean by the move to fourier space?

make_coordinates should not call something called _fftfreqs though… I made _fftfreqs private temporarily to hide this issue.

There is also an annoying detail with jax_finufft, where I have been needing to fftshift in real space after computing the inverse FFT of a projection image. This could probably be fixed though. This is why initially I had my own FFT conventions in cryojax.utils.fft.

mjo22:
On this topic:

XLA turns ``jnp.exp(jnp.log(x))`` -> ``x`` upon jit compilation.

Does it do the same for ``jnp.fft.fftn(jnp.fft.ifftn(x))``? This would make part of this issue a non-issue, as there is some control flow that would not need to be written.

mjo22:
Closing with #105

Issue #54, opened 2023-12-16
Option to return images in fourier space


mjo22:
Add an option in ImagePipeline.render and sample to return images in fourier space. See if we can cut down on on the number of fourier transforms back and forth

mjo22:
See #68. This should be done in a way so that no FFTs are used to get an image in fourier space for fourier slice theorem

PR #100, opened 2023-12-30
Simplify the ImagePipeline and refactor the Distribution class

### LINKED ISSUE #12: Simulate multiple specimen in a field of view (Micrograph class?)
`Image` models should compute the image from multiple `Specimen` in a field of view.

It is unclear whether this functionality belongs in the `Image` class, or if it is a larger project for a new `Micrograph` class.

mjo22:
I've been working on some functionality for my work, where I sum together projection images when the `Ensemble` class has a batch dimension in the `Pose` and/or `Conformation`. 

I was doing this a little awkwardly, where I was allowing for the `ImagePipeline` to take in this `Assembly` class I made. However, this was not good because I really don't want to build in core functionality for an individual person's work.

So I instead made what's called a `SuperpositionPipeline`, which is a child of the `ImagePipeline`. It computes images by vmapping over poses and/or conformations and summing together the stack of images (which is much more general than my project! can be used to close #12). This also required that the `Distribution` class to no longer inherit from the `ImagePipeline`. Instead, it takes an `ImagePipeline` or child classes as input. I have also addressed #52 and renamed `GaussianImage`.


mjo22:
Additionally made progress in #68 and #54

mjo22:
Also made some miscellaneous updates… the log_probability function was not particularly compatible with the use of filters. Will be doing testing with the WhiteningFilter in the coming months. This should fix things.

PR #102, opened 2024-01-05
fourier slice insertion

geoffwoollard:
TODO

- [x] find a home for `insert_slice`. Does it belong in `cryojax.simulator.scattering._fourier_slice`? Perhaps not in scattering but in inference 
- [ ] jaxify pure numpy code in cryojax.utils.interpolate.diff, fill_vec, interp_vec 
- [x] add in ctf deconvolution (wiener filtering)
- [ ] demo reconstruction in notebook. a good time for code to sample rotations uniformly

Notes
Code taken from https://github.com/compSPI/reconstructSPI/blob/insert_slice_trilinear/reconstructSPI/iterative_refinement/interpolate.py and https://github.com/compSPI/reconstructSPI/blob/insert_slice_trilinear/reconstructSPI/iterative_refinement/expectation_maximization.py#L643 I wrote most of this, but there's an attribution trail to be added.


geoffwoollard:
Thinking more about what belongs in inference.... eventually gradient based methods, sampling / monte carlo methods. The Fourier back projection is a "direct solution" technique like common lines. Inverse problem sometimes have methods like this, where the measurements can be "directly" combined.

So perhaps this means that the Fourier insertion and back projection is less "inference". The pose and other parameters are inferred...

Maybe a "solve" or "reconstruct" category makes sense for back projection. Or just a backproject.py itself.

geoffwoollard:
In a sense reconstruction by Fourier slice insertion is analogous to Weiner filtering.

As part of the forward model, the CTF multiplies an image in reciprocal space. This multiplication in the forward model can be "undone" by wiener filtering, with care taken for the zeros in the CTF.

Analogously, as part of the forward model, the projection operator slices the volume in reciprocal space. This slicing can be "undone" by Fourier insertion, with care taken for "off grid" values with (linear) interpolation.

It is not always the case that we can "undo" operations in the forward model in a straight forward way, as information may be lost in a way where we don't know how to recover it sufficiently. Summing on the image direction of the volume in primal space, at first, seems like we can't fully "undo" it. However the Fourier slice theorem gives shows us that we can use the information from the projection, take the FT (no loss of info), and then insert it in a special mathematical space (reciprocal space) that preserves the information in this sense: we can insert many slices in the reciprocal volume, and then do a lossless iFT back to primal space, and recover the volume ... meaningful convergence (one could work out bounds for particular cases).

src/cryojax/simulator/scattering/_fourier_slice.py
```diff
@@ -115,3 +118,52 @@ def extract_slice(
...
+    
+    xyz = jnp.zeros((3,N**2))
+    rotation = pose.rotation.as_matrix()
+    xyz_central_slice = xyz.at[:2].set(make_coordinates((N,N)).reshape(-1,2).T)
+    xyz_rotated_central_slice = rotation @ xyz_central_slice
+
+    inserted_slice_3d_real, count_3d_real = insert_slice_and_interpolate(slice_f.real, xyz_rotated_central_slice, N)
+    inserted_slice_3d_imag, count_3d_imag = insert_slice_and_interpolate(slice_f.imag, xyz_rotated_central_slice, N)
+
+    volume = inserted_slice_3d_real*count_3d_real + 1j*inserted_slice_3d_imag*count_3d_imag
```
geoffwoollard, line 164: Normalization OK?

mjo22:
@geoffwoollard could you make this a PR to helix? api-update is pretty old now and only isn’t deleted because of the other PRs to it. This is the most updated branch and may be helpful for this. 

Hopefully next week we will get all of these updates merged into main, then we can update this PR to be there instead.

geoffwoollard:
Hmm, not sure why this isn't recapitulating things better. Might be related to back projection vs filtered back projection. 

7035d212-f5ee-4969-bc39-b8051f8d0506
a97fb598-5ba9-45b2-8666-bc094d6d3bc7
8c4bf752-9981-4b9c-b9bb-494f2722ba1b
02df8733-02f9-492b-b22b-9c74f66d657b

geoffwoollard:
Hmmm, filtering with the absolute frequency (ram-lak) filter isn't seeming to help...

There's plenty of youtube videos about this.
https://youtu.be/qfAS8seVYvU?si=ukAHZ1c_6u-FqgOq&t=1431
https://youtu.be/ZgcD4C-4u0Q?si=Or7ejXSCdmDQiYHz&t=3253

3a6a6c25-c05b-43e1-873b-567c24bebce8
2d4cb514-bef8-419b-877e-e24eb4d89b26
7d8ba9c4-087a-48e0-b921-68e056f985e3

geoffwoollard:
- [ ] find jax native deconvolution to act as wiener filter
- [ ] use where in interp_vec for boolean good indeces
- [ ] loop in slice insertion with fori_loop: https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#fori-loop
- [ ] rename inference to reconstruct, and have reconstruct/filteredbackprojection.py

mjo22:
See also the new updates in helix to the Filter class… maybe it would make sense to unify the Filter base class and with the idea of a DeconvolvingFilter? Or create a subclass that acts as an abstract base class for a DeconvolvingFilter? Then at some point we can think of where to move Filter and Mask, outside of cryojax.simulator.

geoffwoollard:
I merged in wiener_filter, which was branched off of helix.

geoffwoollard:
@mjo22 comments appreciated.

Did some refactoring. It's still not jaxy, but let's settle on a some high level reconstruction design conventions before I do that. Specifically, what do you think of this naive type signature (just to get the conversation going)?

```
def filtered_backprojection(
    deconolved_images_f: [ComplexImage, ...],
    poses: [Pose, ...],
    to_real: bool = True,
) -> Union[Volume, ComplexVolume]:
```

Since there are not dataclsses like "particle stack" I just pass in two lists. But it might be a good idea to use a dataclass that can tie together the latents (per particle ctf, pose; global params like kv, pixel size) and observes (image) and intermediates (deconvolved image)

tests/test_backproject.py
```diff
@@ -0,0 +1,131 @@
...
+    fbp_vol_r /= fbp_vol_r.std()
+
+    gt_vol_r = ifftn(density.weights).real
+    gt_vol_r -= gt_vol_r.mean()
+    gt_vol_r /= gt_vol_r.std()
+
+    rel_resid = jnp.abs((gt_vol_r - fbp_vol_r)).mean() / gt_vol_r.std()
+    print("residual_ratio", rel_resid)
+    residual_ratio_tol = 0.9
+    assert rel_resid < residual_ratio_tol
```
geoffwoollard, line 127: This test is sort of working, it's just inefficient for now since the code is not jaxy

| n_slices |  residual_ratio | test time|
|-|-|-|
|30 | 0.84| seconds |
|100 | 0.77| less than 1 min|
|1000|0.46| several min|

mjo22:
Hi, taking a look at this now. I agree the dataclass would be nice, but we should definitely just get this merged before. Things look great.

I’ve made significant updates to the main branch, so we can get things merged once resolving conflicts.

For resolving conflicts,

1) The addition of the `cryojax.image.operators` module. This houses the `Filter` class. Note also I now use significantly the `FourierOperator` class. This is basically a function in fourier space that is computed with `__call__`. A `Filter` on the other hand computes an array upon `__init__` and stores it as a buffer. Then, `__call__` applies this buffer to an image. The advantage of the `FourierOperator` is you can do something like gradient descent on the parameters, and the advantage to the `Filter` is you can compute an array once to store, which you could then pass into some kind of, say, training loop and not have to worry about doing repetitive computation. I’m not sure which class is more appropriate for the `WeinerFilter`, but wanted you alert you to this.
2) I independently implemented cropping in 3D (note the rename) cryojax.image._edges.py, but a little more generally to work with arrays with leading dimensions, so we can keep my version (assuming it is correct).
3) there is a new class called `CTF`. This belongs to the `CTFOptics` class, which also has an envelope function. 
4) I’ve started on the `cryojax.inference` module, adding a `Distribution` class. I think there may be a conflict there but that should be easily resolved.

geoffwoollard:
Thanks for those comments, especially how to use vmap.

Still wondering about the overall scale: how the FFTs and iFFTs are scaled, and how to deal with the DC component going to zero in the ram-lak (absolute frequency) filter for "filtered" back projection.

mjo22:
I'm not familiar with the latter, but I've removed my ad-hoc modification to the fourier amplitudes (the `1 / sqrt(N_modes)` factor). So this shouldn't be a problem with the new updates!

geoffwoollard:
Talked with Niko on Friday, and there's a way to filter by using the accumulated counts from the slices themselves. 

I'm using a theoretical filter (abs freq), but the counts from each slice being inserted (which I already compute) can be passed through and accumulated. You get the counts from just inserting an image of all ones. There's a "count volume" and a "slice inserted volume", and they can be divided out at the end.

Niko suggested two papers to look at, for how this plays with the CTF.

I think it's worth taking the time to set up the flexibility for what comes from data and what doesn't. It's not that we have to code up reconstruction in a specific way, but just give enough flexibility for different options to be possible with the code.

* Sindelar, C. V., & Grigorieff, N. (2011). An adaptation of the Wiener filter suitable for analyzing images of isolated single particles. Journal of Structural Biology, 176(1), 60–74. http://doi.org/10.1016/j.jsb.2011.06.010
* Sindelar, C. V., & Grigorieff, N. (2012). Optimal noise reduction in 3D reconstructions of single particles using a volume-normalized filter. Journal of Structural Biology, 180(1), 26–38. http://doi.org/10.1016/j.jsb.2012.05.005

mjo22:
A note on the PR: I think it’s probably more appropriate for the interpolation functionality to be moved to private methods in the `cryojax.reconstruction` submodule, rather than being in `cryojax.image`. `cryojax.image` has become focused on providing generally useful routines for working with images and volumes, whereas these interpolation routines seem to be tailored to the specific task of fourier slice insertion (correct me if I’m wrong on this).

(I know I’m contradicting what I recommended to you at an earlier time… thank you for being patient with all of the many API updates)

mjo22:
over time has been more clear that functionality like this should be downstream; cryojax is more for forward model

PR #106, opened 2024-01-10
Refactor the Ensemble, ElectronDensity, and Conformation

mjo22:
The ``_is_stacked`` interface is not great in the ``ElectronDensity``, so in my opinion there is a big argument for just building generic batching utilities into cryojax. This is a common thing to do in popular jax libraries, such as ``numpyro`` and ``distrax``. The way we would do this is build functionality into ``cryojax.Module``. I've started on a rough draft of what this may look like.

These updates allow us to make sure that the code does not unexpectedly break when objects have batch dimensions. Right now, we would have to go through the code and generalize function calls to make use of the new ``Module.batch_axes`` parameter. For example, FFT calls should not FFT on the ``batch_axes``.

Also ``Module.batch_axes`` is just a rough draft. It also currently only supports ``batch_axes = (0,)``. Need to make ``__getitem__`` more generally support index slicing, and maybe even fancy slicing.

There is also an argument for making every class in ``cryojax`` optionally implement their own ``__getitem__``. Either this or they could override the ``cryojax.Module.__getitem__`` (for example ``VoxelCloud`` would need to do this).

This is a rough draft of what things could look like, but I think it's the right direction.


mjo22:
In the current paradigm we assume that the batching in just in the leading axes, so changing ``batch_axes`` to ``n_batch_dims``. Don't love the name so far

mjo22:
It would also be nice to implement fancy indexing, and support multiple batch dimensions. Need to do this for ``__getitem__``.

mjo22:
There is also maybe argument for always giving ``Ensemble`` an electron density with a batch dimension.

See here for potential inspiration:
https://github.com/google-deepmind/distrax/blob/master/distrax/_src/distributions/mixture_same_family.py

For this, we need to make sure this isn't more cumbersome for the user if they want to simply simulate an image of one electron density. Perhaps this means the old ``Specimen`` class should be revamped.

mjo22:
Merging for now---things were better than before. The solution right now is that using the argument ElectronDensity.n_stacked_dims is ONLY for use with Ensemble.conformation. This is not to be used for general vmapping. The user should handle a case like this in their scripts.

Issue #108, opened 2024-01-11
wiener filter


geoffwoollard:
I branched wiener_filter off of helix.

It makes sense to me that wiener_filter is applied to images that come out of image_pipeline, with a detector.

I thought about where to put the filter to undo the noisy image... it should match up with the ctf already in the image pipeline (padding, freq, etc). 

Here are some ideas:
1.  it's just a function that acts on a plain jax array, or something that can take in a ImagePipeline
2. ImagePipeline (a) or Instrument (b) should itself take a filter that undoes the affect of the instrument
3. We need a new category like (particle stack) that will be used for reconstruction. Using the categories for the simulator is awkward since the simulator is a forward model that goes in one direction, while reconstruction is an "inverse model" that goes in the other direction.

PR #109, opened 2024-01-11
wiener filter

geoffwoollard:
See issue https://github.com/mjo22/cryojax/issues/108

- [x] decide where to put the wiener filter
- [ ] make some tests
  - [x] wiener filter with noise level 0 is same as dividing by ctf (up to atol)
  - [ ] convolve and then deconvolve (with WienerFIlter) image with ctf (with small amount of noise) and show get back similar values.

ae013efb-6bb6-424d-b4d3-917b44a3737b
0c66d72b-d586-42a3-b0dd-60ecc00bce92
24958995-b030-4503-8032-883b2b6bcdd1
383118c4-34a0-411b-b79b-e82e500b9a18
992b670d-2618-4159-adcb-10d4ad844d1d


geoffwoollard:
@mjo22 if you explain to me a bit more about how the testing works in terms of `@pytest.mark.parametrize("model", ["noisy_model"])`, then I can make tests for

- [ ] convolve and then deconvolve (with WienerFIlter) image with ctf (with small amount of noise) and show get back similar values.

mjo22:
For this PR, I think we should keep the WeinerFilter in the reconstruct branch and create a new module for it. For now, you can import the Filter base class from the simulator and subclass it in the reconstruct module. Later, we can figure out where we should move the Filters (maybe we should move into cryojax.core).

I’m also not sure if the way I’ve written the Filter class will be compatible with your use case. If you want to subclass, you could run something like

```python
from cryojax.simulator import Filter

class WeinerFilter(Filter):

    def __init__(ctf: Realmage, noise_level = …):
        self.filter = …  # Set the filter with the CTF
```

and call it with

```python
from cryojax.simulator import CTFOptics

frequency_grid = make_frequencies(…)
optics = CTFOptics(…)
ctf = optics(frequency_grid)
filter = WeinerFilter(ctf)
```

Not sure if this achieves what you need, but this is the simplest I can think of. Also note that the snatch with the Filter is it cannot contain model parameters because I compute the filter at initialization, not runtime. This is because for the simulator, we only want to compute it once and store it as a buffer.

mjo22:
> @mjo22 if you explain to me a bit more about how the testing works in terms of [/mnt/home/gwoollard/repos/cryojax/tests/test_filters_and_masks.py](https://github.com/mjo22/cryojax/blob/6151ab222ecc3857264c15a3ceaed39bfbc389bf/tests/test_filters_and_masks.py#L8):`@pytest.mark.parametrize("model", ["noisy_model"])`, then I can make tests for
> 
> * [ ]  convolve and then deconvolve (with WienerFIlter) image with ctf (with small amount of noise) and show get back similar values.

For this, read the conftest.py. This is a detail of how pytest works—it knows to look at the conftest.py.

geoffwoollard:
@mjo22 I refactored as you said and put in a test. It's failing in the github workflow, but is passing locally with 

```
(cryojax) (base) [gwoollard@workergpu054 cryojax]$ pytest tests/test_backproject.py
============================================================== test session starts ==============================================================
platform linux -- Python 3.10.13, pytest-7.4.4, pluggy-1.3.0
rootdir: /mnt/home/gwoollard/repos/cryojax
plugins: jaxtyping-0.2.24, typeguard-2.13.3
collected 2 items                                                                                                                               

tests/test_backproject.py ..                                                                                                              [100%]

=============================================================== 2 passed in 1.57s ===============================================================
```

geoffwoollard:
Closing since so much of the code base has changed.

PR #127, opened 2024-01-16
Atom loading refactor and Added ability to load mdtraj trajectories

ehthiede:
Loading atoms have been refactored so that we can now vmap over xyz positions stored in numpy arrays.  Relatedly, I've added the ability to load molecular dynamics trajectories into densities via the `mdtraj` library.


mjo22 (approved):
Looks awesome!

Issue #133, opened 2024-01-21
radial_average


geoffwoollard:
I wanted to whiten a particle stack and was hoping to use the `WhiteningFilter`.

I did some sanity checks on random gaussian noise (already white), and suspect there is a bug in `radial_average`.

When I radially average the spectra, I don't get back a flat filter. Of course there is some sampling noise here since the image (here 64 x 64) is finite. But even in expectation (1000 samples), it's not flat.

```python
shape = (64,64)
k_min, k_max = 0, 0.5
pixel_size=1
k_step = 1.0 / (pixel_size * max(*power.shape))
bins = jnp.arange(k_min, k_max, k_step)
frequency_grid = np.fft.fftshift(make_frequencies(shape, grid_spacing=1, half_space=False))
radial_frequency_grid = jnp.linalg.norm(frequency_grid, axis=-1)

freqs = frequency_grid
interpolating_radial_frequency_grid = jnp.linalg.norm(freqs, axis=-1)

n_samples=10000
spectra = np.empty((n_samples,*shape))
for idx, seed in enumerate(range(n_samples)):
    np.random.seed(seed)
    image_r = np.random.normal(np.zeros(shape), 1)
    image_f = np.fft.fftshift(np.fft.fftn(image_r))

    power = (image_f * jnp.conjugate(image_f)).real
    spectrum = radial_average(
        power, radial_frequency_grid, bins, interpolating_radial_frequency_grid
    )
    spectra[idx] = spectrum
```

```
plt.imshow(spectra.mean(0))
plt.colorbar()
plt.title(f'Expectation of radially averaged power of random gaussian noise \n n_samples={n_samples}')
```
![0b82604a-9683-47e6-bbae-9117ec11f6bf](https://github.com/mjo22/cryojax/assets/4852692/9fc475ac-471a-46f9-a381-e660a028936e)

```
plt.imshow(spectra.std(0))
plt.colorbar()
plt.title(f'Std of radially averaged power of random gaussian noise \n n_samples={n_samples}')
```
![c13c2f64-20e5-45ec-9861-78c138112918](https://github.com/mjo22/cryojax/assets/4852692/87d2441f-a8db-4b89-a407-fbb26289e6f7)

Using code from commit b39bd0de800692503d806844c45fee76ceaf18a0

geoffwoollard:
Here is just radially averaged gaussian noise

[...]

PR #134, opened 2024-01-21
radial average patch. should be using jax.numpy.take to put on grid

mjo22:


mjo22:
Adding some more fixes for #133. need to safely treat out of bounds indexing when using digitize.

Issue #116, opened 2024-01-13
documentation for relion convention


geoffwoollard:
@mjo22 I think I figured out the correspondence of cryojax to Relion:

Note: 
This matched with relion star file conventions, where data has a star file loaded in:
shot_info_view_phi -> _rlnAngleRot
shot_info_view_theta -> _rlnAngleTilt
shot_info_view_psi -> _rlnAnglePsi
shot_info_offset_x -> rlnOriginXAngst  (or info['rlnOriginX'] * pixel_size_angstrom)

```
volume_t = jnp.transpose(volume, axes=[2, 1, 0]) # NOTE TRANSPOSE
density = cs.VoxelGrid.from_density_grid(volume_t)
shape=density.weights.shape[:2]
pad_scale = 1

pixel_size = data['pixel_size']
manager = cs.ImageManager(shape=shape, pad_scale=pad_scale)
scattering = cs.FourierSliceExtract(manager, pixel_size=pixel_size)

pose = cs.EulerPose(view_phi=data['shot_info_view_phi'][idx],
                    view_theta=data['shot_info_view_theta'][idx],
                    view_psi=data['shot_info_view_psi'][idx],
                    offset_x=data['shot_info_offset_x'][idx],
                    offset_y=data['shot_info_offset_y'][idx],
                    convention='zxz', # NOTE CONVENTION IS NOT DEFAULT OF CRYOJAX
                    )
ensemble = cs.Ensemble(density=density, pose=pose)

optics = cs.CTFOptics(defocus_u=data['shot_info_defocus_u'][idx],
                    defocus_v=data['shot_info_defocus_v'][idx],
                    defocus_angle=data['shot_info_defocus_angle'][idx],
                    voltage=data['shot_info_voltage'][idx],
                    spherical_aberration=data['shot_info_spherical_aberration'][idx],
                    amplitude_contrast=data['shot_info_amplitude_contrast_ratio'][idx],
                    phase_shift=data['shot_info_phase_shift'][idx],
)
instrument_od = cs.Instrument(optics=optics)
pipeline = cs.ImagePipeline(scattering=scattering, instrument=instrument_od, ensemble=ensemble)
image = pipeline.render()

image = image[::-1] # NOTE MIRRORING
```

The image associated with the `idx` in the relion star file will match image

mjo22:
[...]

PR #150, opened 2024-02-01
Rotation logic is more clear. Now, AbstractPose rotation parameters by default define a rotation of the real space volume. Before, they were with respect to a fourier space volume.

mjo22:
Found a bug in the AbstractAssembly that was computing the inverse rotation on accident of the subunit positions. There were accidental transposes lying everywhere in the Helix class. The reason for the bug was unclear naming in the AbstractPose.rotate_coordinates function.

This lead to the larger change that now, the AbstractPose rotation defines a rotation of the **real space** volume. Before, the rotation was sneakily with respect to the fourier space volume.

### DIFF
--- a/src/cryojax/coordinates/__init__.py
+++ b/src/cryojax/coordinates/__init__.py
@@ -1,5 +1,5 @@
 from ._coordinates import (
-    is_not_coordinate_array as is_not_coordinate_array,
+    is_not_coordinates as is_not_coordinates,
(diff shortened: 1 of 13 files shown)


mjo22:
@geoffwoollard I wonder if this could have also been a reason for the discrepancy in #116? Are relion euler angles with respect to real or fourier volume?

mjo22:
helical assembly tests still not passing

PR #155, opened 2024-02-06
Get rid of file loaders in the electron density

mjo22:
I think it's best all file loaders are removed from the electron density. This is for three main reasons

1. I think it is bad practice to hide user interaction with `cryojax.io`.
2. Relatedly, the file loaders hide what exactly should be passed to the constructors.
3. Finally, I am going to start maintaining that all functions in `cryojax` objects must work with `jax.jit`. Otherwise, users could get confused and it would require additional documentation on our part.

With this PR, the only custom constructors are `AbstractVoxels.from_density_grid` and  `AbstractVoxels.from_atoms`. I also propose getting rid of `AbstractVoxels.from_trajectory` in favor of vmapping over `AbstractVoxels.from_atoms`. `AbstractVoxels.from_trajectory` presents some issues with typing, and it requires documenting what exact fields are vmapped over. I've updated the test in `test_voxels_from_atoms.py` to show how the `from_trajectory` behavior can easily be replicated.

@ehthiede, what do you think? You have suggested this in the past, but at the time I wasn't on board...

### DIFF
--- a/docs/api/simulator/electron_density.md
+++ b/docs/api/simulator/electron_density.md
@@ -17,6 +17,7 @@
             members:
                 - is_real
                 - shape
+                - rotate_to_pose
                 - from_density_grid
                 - from_atoms
 
@@ -44,6 +45,8 @@
                 - voxel_size
                 - frequency_slice
                 - frequency_slice_in_angstroms
+                - from_density_grid
+                - from_atoms
 
 ---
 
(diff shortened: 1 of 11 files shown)


ehthiede:
I'm down with this!

ehthiede (approved):

mjo22:
Awesome

PR #160, opened 2024-02-08
added relion star file and mrcs file. 10 images with 48x48 images

Miro-Astore:
pixel size is 4.88 A. Drawn from reconstruction of EMPIAR 10059. 4x downsampled.


mjo22:
awesome, will mess around with the stack and merge sometime next week. could you move this PR to the new starfile branch?

PR #162, opened 2024-02-11
First pass at simulating images in physical units

mjo22:
Gave a first pass at simulating images in electron counts. I'm not too worried about it being exactly correct for now, but this PR includes API changes. Renaming was necessary to get closer to something that is a physical model.

1. API Changes
 - The biggest API change is that the "electron density" has been renamed to the "scattering potential". This includes `AbstractElectronDensity` -> `AbstractScatteringPotential` and local variable names `density` -> `potential`. Really this should be something like `phase_shifts_due_to_potential`, but we can document that we are assuming a non-dimensionalized potential
 - For clarity, I've also changed `AbstractScatteringMethod` -> `AbstractPotentialIntegrator`. Similarly, variables names `scattering` -> `integrator`. The idea of a "scattering method" I think is unclear. This is really just a choice of integration method onto the exit plane, for a given scattering potential representation.
 - Have also renamed `CTFOptics` -> `WeakPhaseOptics`. Wasn't sure for a while how this should be renamed, but I've made clear now that an `AbstractOptics` model takes in a potential in the exit plane and propagates it to the detector plane. When the weak phase approximation is made, linear image formation allows us to directly apply the CTF to the potential to get the contrast. If weak phase is not made, an optics model could compute a wavefunction from the potential and instead propagate this to the detector plane.
 - Renamed some parameters in the `CTF` to make them much more descriptive.
2. The `AbstractExposure` has been removed and replaced with `ElectronDose`. Right now, `ElectronDose` just contains a single parameter `electrons_per_angstrom_squared`. One could image in the future, this object could be responsible for modeling radiation damage, taking multiple exposures, etc.
3. The `AbstractDetector` has been given a first pass at being a real detector model. This includes a new `PoissonDetector` and `GaussianDetector`, and a model for the DQE through the `AbstractDQE` interface. (NOTE: The `GaussianDetector` is the gaussian limit of the `PoissonDetector`. If the user wants a noise model that is just image + std * white_noise, the recommendation will be to use a `cryojax.inference.distributions.AbstractDistribution` object).
4. Renaming in the image pipeline internals to make clear what exactly is being computed at each step (for example, `image_at_exit_plane` -> `potential_at_exit_plane`)

I think things are good enough to merge for the updates (only if this would be okay for the other PRs), but---using the cisTEM scattering potential---I am not sure yet if the images I am simulating are actually in electron counts. For this, will have to make sure the projected scattering potential I'm computing is actually a phase contrast (i.e. that I'm not missing any units). Things will not yet be working with in the atoms -> scattering potential functionality native to cryojax.

A tricky part in this is correctly converting the contrast to the squared wavefunction. The Hawkes and Kasper wave optics book defines 

$$C(\mathbf{x}) = \frac{1 - |\psi(\mathbf{x})|^2}{1 + |\psi(\mathbf{x})|^2}$$

where $C$ is the contrast and $\psi$ is the wavefunction. This is solved by

$$|\psi(\mathbf{x})|^2 = \frac{1 - C(\mathbf{x})}{1 + C(\mathbf{x})} \approx  1 - 2 C(\mathbf{x}),$$

where it is assumed $C << 1$. In fourier space this is

$$\mathcal{F}\[ |\psi(\mathbf{x})|^2\](\mathbf{q}) = \delta(\mathbf{q}) - 2 C(\mathbf{q})$$


mjo22:
The thing I don't understand is that, in the current scheme of things, the poisson rate is set to be equal to the `electrons_per_pixel` for an image with nothing in it (vaccum). This allows for electron counts in vaccum greater than the `electrons_per_pixel`. This doesn't seem physical.

Now, of course I'm doing this all for a DQE = 1 (I think this means an ideal detector). Still, it doesn't seem to me like this should happen. Maybe I need to actually model the DQE to fix this.

geoffwoollard:
> The thing I don't understand is that, in the current scheme of things, the poisson rate is set to be equal to the `electrons_per_pixel` for an image with nothing in it (vaccum). This allows for electron counts in vaccum greater than the `electrons_per_pixel`. This doesn't seem physical.
> 
> Now, of course I'm doing this all for a DQE = 1 (I think this means an ideal detector). Still, it doesn't seem to me like this should happen. Maybe I need to actually model the DQE to fix this.

See the paper Loh, N. D. (2014). A minimal view of single-particle imaging with X-ray lasers. Philosophical Transactions of the Royal Society B: Biological Sciences, 369(1647), 20130328. http://doi.org/10.1098/rstb.2013.0328

See especially section "(a) Separation parameter", p. 3 and Figure 3.

In the limit of large electrons_per_pixel the effective variance of the Poisson operator goes to zero (Gaussian becomes a delta function). 

However, for low electrons_per_pixel, the Poisson operator can with some probability give large counts. There is no conservation of noise for one stochastic event, only in expectation.

src/cryojax/simulator/_detector.py
```diff
@@ -3,117 +3,161 @@
...
     @override
     def sample(
-        self,
-        key: PRNGKeyArray,
-        image: RealImage,
-        coords_or_freqs: ImageCoords,
+        self, key: PRNGKeyArray, expected_electron_events: RealImage
     ) -> RealImage:
-        return jr.poisson(key, image).astype(float)
+        return jr.poisson(key, expected_electron_events).astype(float)
```
geoffwoollard, line 163: you could illustrate a case of having `expected_electron_events` as `
dose*image + background` in a notebook.

Doing inference in this model could have dose and background as a global scalar parameters, and image as a per measurement parameter  of shape (n_pix,n_pix).

src/cryojax/simulator/_optics.py
```diff
@@ -17,214 +16,176 @@
...
+              in degrees or radians.
     """
 
     defocus_u: Real_ = field(default=10000.0, converter=jnp.asarray)
     defocus_v: Real_ = field(default=10000.0, converter=jnp.asarray)
-    defocus_angle: Real_ = field(default=0.0, converter=jnp.asarray)
-    voltage: Real_ = field(default=300.0, converter=jnp.asarray)
-    spherical_aberration: Real_ = field(default=2.7, converter=jnp.asarray)
-    amplitude_contrast: Real_ = field(default=0.1, converter=jnp.asarray)
+    astigmatism_angle: Real_ = field(default=0.0, converter=jnp.asarray)
```
geoffwoollard, line 48: could have units here too: `astigmatism_angle_in_rad/deg`

src/cryojax/simulator/_optics.py
```diff
@@ -17,214 +16,176 @@
...
+
+    `amplitude_contrast_ratio`: The amplitude contrast ratio.
+
+    `phase_shift`: The additional phase shift.
+
+    `degrees`: Whether or not the `astigmatism_angle` and `phase_shift` are given
+              in degrees or radians.
     """
 
     defocus_u: Real_ = field(default=10000.0, converter=jnp.asarray)
```
geoffwoollard, line 46: `defocus_u/v_in_angstroms`

src/cryojax/simulator/_optics.py
```diff
@@ -17,214 +16,176 @@
...
-            self.voltage,
-            self.spherical_aberration,
-            self.amplitude_contrast,
-            self.phase_shift,
-            degrees=self.degrees,
+            self.defocus_u + jnp.asarray(defocus_offset),
+            self.defocus_v + jnp.asarray(defocus_offset),
+            astigmatism_angle,
+            self.wavelength_in_angstroms,
+            self.spherical_aberration_in_mm * 1e7,  # mm to Angstroms
```
geoffwoollard, line 73: it's a small thing, but I would document the comment as code:

spherical_aberration_in_angstroms = self.spherical_aberration_in_mm*1e7

and then use spherical_aberration_in_angstroms in the call

geoffwoollard: comments are easier to loose track of

geoffwoollard (changes requested):
Do you have some tests for the Poisson noise --> Gaussian at high dose? I would add those in with enough numerical tolerance that they agree. It will also be good documentation of the relationship between Poisson dose and Gaussian variance.

Some tests could be in expectation, so that the variance agrees numerically. E.g. simulate a bunch of images from the same base potential, and then check that the SNR is the same in expectation.

geoffwoollard:
> A tricky part in this is correctly converting the contrast to the squared wavefunction. The Hawkes and Kasper wave optics book defines
> 
> C(x)=1−|ψ(x)|21+|ψ(x)|2
> 
> where C is the contrast and ψ is the wavefunction. This is solved by
> 
> |ψ(x)|2=1−C(x)1+C(x)≈1−2C(x),
> 
> where it is assumed C<<1. In fourier space this is

This is different from what I'm familiar with. See "2.2. Electron wave propagation" in Vulović, M., Ravelli, R. B. G., van Vliet, L. J., Koster, A. J., Lazić, I., Lücken, U., … Rieger, B. (2013). Image formation modeling in cryo-electron microscopy. Journal of Structural Biology, 183(1), 19–32. http://doi.org/10.1016/j.jsb.2013.05.008

src/cryojax/simulator/_detector.py
mjo22: In this case, what is background? Is this to account for the possibility of not having the right global mean?

src/cryojax/simulator/_optics.py
mjo22: I’ve thought about this also… since I’ve added the option everywhere in the package to use degrees or radians with an optional `degrees` boolean, I wasn’t sure a good solution to this

src/cryojax/simulator/_optics.py
mjo22: Also wasn’t sure about this… was tempted to keep things similar to relion and CTFFIND, where I think defocus is always assumed to be measured in angstroms?

src/cryojax/simulator/_optics.py
mjo22: Good point yes

mjo22:
Re detector test: yes this would be a good to add… will add that as something to do. right now do not have testing for the detector. 

Re this paper and the squared wavefunction: I have been reading this paper yes… the expression I wrote down comes out of the derivation of linear image formation theory, which is not what this paper uses. In linear theory, the main object one works with is the contrast, not the wavefunction. However, one can derive how the contrast is related to the squared wavefunction.

mjo22:
Added test for the `GaussianDetector`, making sure it's really the gaussian limit of `PoissonDetector`

src/cryojax/simulator/_optics.py
geoffwoollard: Yeah I saw the degrees bool after I wrote this comment. 

Besides checking by assertion, I'm not sure how this could be verified. Might have to rely on user to not introduce bugs.

src/cryojax/io/_mrc.py
```diff
@@ -3,66 +3,173 @@
...
+
+
+def _read_array_from_mrc(
+    filename: str | Path, get_spacing: bool, mmap: bool
+) -> Float[np.ndarray, "..."] | tuple[Float[np.ndarray, "..."], float]:
     # Read MRC
-    with mrcfile.open(filename) as mrc:
-        data = np.asarray(mrc.data, dtype=float)
+    open = mrcfile.mmap if mmap else mrcfile.open
+    with open(filename, mode="r") as mrc:
```
geoffwoollard, line 131: I would include options for `permissive=True` - since it's not uncommon with older mrc files in the EMDB that they contain bad header data where mrcfile.open will fail.

See https://mrcfile.readthedocs.io/en/stable/usage_guide.html

geoffwoollard (commented):
Nice test!

Yeah I think you're 98% there.

I would take a look at the few comments, decide if you want to do anything with my suggestions, and then merge.

src/cryojax/simulator/_optics.py
mjo22: looked more into this, looks like ctffind has both “defocus1_in_angstroms” and “defocus1” (which may not be in angstroms? a little confused about this). but changed names to defocus_u/v_in_angstroms

src/cryojax/io/_mrc.py
mjo22: Cool, this is really good to know. Will add this option. I had to do a merge while working on different branches, but most io updates will be coming from the starfile branch…

The mrc loading I’ve been struggling with a bit because I’ve been using different axis conventions than mrc format… at this point it might be good to switch to z, y, x axis ordering (rather than x, y, z like I’ve been using), but it’s a real pain to convert. Will be working on starfile compatibility in the coming weeks so it should become clear what the right thing to do is.

src/cryojax/simulator/_optics.py
mjo22: Yes, either this or remove the option all together and stick to degrees… not totally sure but we can stick to this for now.

mjo22:
Thank you for reviewing!

PR #178, opened 2024-03-14
Reorganization of the ImagePipeline

mjo22:
This PR is an API change that

1. Moves the `ImageConfig` directly to the `AbstractPipeline`. This makes the `AbstractPipeline` much more flexible and gives it a clear abstract interface. The `ImageConfig` gets handed to a lot of simulator objects under the hood, so this change makes the simulator internals much cleaner I think. Before, it belonged to the `AbstractPotentialIntegrator` and then was awkwardly handed to everything else
2. The `AbstractPotentialIntegrator` is then moved to the `AbstractSpecimen`. This is a great change also because now the potential and its integrator are both handed to the same object. This is much more intuitive.

This PR also includes experimental features with testing to come. It also has some fixes and updates in the now dropped jaxlie dependency.

This PR will mark a 0.2.1 release.


src/cryojax/core/_errors.py
```diff
@@ -9,13 +9,27 @@
...
+
+
+def error_if_zero(x: ArrayLike) -> Array:
+    x = jnp.asarray(x)
+    return eqx.error_if(
+        x, jnp.isclose(x, 0.0), "A non-zero quantity was found to be zero!"
+    )
 
 
 def error_if_not_fractional(x: ArrayLike) -> Array:
```
geoffwoollard, line 29: maybe positive fractional?

src/cryojax/core/_filtered_transformations.py
```diff
@@ -9,7 +10,7 @@
 from typing import Callable, Union, Any, Optional, Hashable
 
 
-def filter_grad(
+def filter_grad_with_spec(
```
geoffwoollard, line 13: do you want to give a bit of orientation to these functions with a doctstring?

src/cryojax/inference/__init__.py
```diff
@@ -1,2 +1,2 @@
 from . import distributions as distributions
-from . import parameters as parameters
+from . import transforms as transforms
```
geoffwoollard, line 2: why can't you just do `from . import transforms`? Why do you need the `as`?

src/cryojax/inference/transforms/_lie_group_transforms.py
```diff
@@ -0,0 +1,131 @@
+"""
+Implementation of transformations in local tangent spaces of lie groups.
+This is a re-implementation of `jaxlie.manifold`.
+
+This was necessary because `cryojax` has its own pattern for `jax`
+transformations, using `equinox`. The `jaxlie.manifold.grad` and
+`jaxlie.manifold.value_and_grad` are not compatible with the `cryojax`
+wrappers for gradient transformations.
```
geoffwoollard, line 8: good to put these comments in

geoffwoollard: You might want to point to specific parts in jaxlie.manifold. Or at least a specific version/commit of jaxlie.

src/cryojax/rotations/_lie_group.py
geoffwoollard, line 1: since you are re-implementing some jaxlie things, perhaps test against the jaxlie version (yes, you then you have test dependencies).

mjo22: good idea!

src/cryojax/simulator/_detector.py
```diff
@@ -43,7 +48,12 @@ def __init__(self):
         self.fraction_detected_electrons = jnp.asarray(1.0)
 
     @override
-    def __call__(self, frequency_grid: ImageCoords) -> Real_:
+    def __call__(
+        self,
+        frequency_grid_maybe_in_angstroms: ImageCoords,
```
geoffwoollard, line 53: maybe??? It's better than nothing, I suppose.

angstroms_or_wavenumbers???

src/cryojax/inference/__init__.py
mjo22: this is an annoying detail in `__init__.py`s where if you do not have the `as`, you get a message like `'transforms' is not accessed` from linters, like Pylance. I'm following the pattern of the equinox ecosystems `__init__.py`s

src/cryojax/simulator/_integrators/_nufft_project.py
```diff
@@ -29,35 +29,35 @@ class NufftProject(AbstractPotentialIntegrator, strict=True):
...
+        config: ImageConfig,
     ) -> ComplexImage:
         """Rasterize image with non-uniform FFTs."""
         if isinstance(potential, RealVoxelGridPotential):
             shape = potential.shape
             fourier_projection = project_with_nufft(
                 potential.real_voxel_grid.ravel(),
                 potential.wrapped_coordinate_grid.get().reshape((math.prod(shape), 3)),
-                self.config.padded_shape,
+                config.padded_shape,
```
geoffwoollard, line 45: I suppose this makes things more functional programming. less states and params stored in self

src/cryojax/core/_filtered_transformations.py
mjo22: yes good idea... I may do this in a future PR because in the next few weeks I'll be writing up some examples using this in practice. right now `cryojax.core` is mostly experimental and is lightweight extensions to `equinox`. I'm excited to write these up---it breaks down `equinox` into its intuitive components and will draw people in to learn more :).

src/cryojax/simulator/_pose.py
```diff
@@ -161,13 +160,6 @@ class EulerAnglePose(AbstractPose, strict=True):
         The sequence of axes over which to apply
         rotation. This is a string of 3 characters
         of x, y, and z. By default, `'zyz'`.
```
geoffwoollard, line 162: Perhaps still include some details about the convention you use (at high level natural language explanation). They can look at your code to see what it does in more detail...

src/cryojax/simulator/_detector.py
mjo22: the reason why I have this is a messy detail I don't quite have resolved in the AbstractFourierOperator class... will document this

src/cryojax/simulator/_integrators/_nufft_project.py
mjo22: its more about the right hierarchy of the pytree I think. the `ImageConfig` needed to be much more privileged because so many objects need it to be passed to them. not sure if this relates to the functional programming aspect of things... i think of this more in the way in which `eqx.Module`s are passed around and operated on by things like `eqx.tree_at` and `jax.tree_util.tree_map`

mjo22:
Thank you for the comments! Merging to stabilize the API, will address these in the next few weeks!

PR #192, opened 2024-03-25
Add dynamic type hints for image shapes at run-time

mjo22:


mjo22:
More run-time type-checker compatibility. pytest is almost happy now. The only test that is not happy is test_voxels_from_atoms.

Issue #196, opened 2024-03-27
Move the accelerating voltage from the `CTF`


mjo22:
For #167, it is looking like we may need to think about where the accelerating voltage / wavelength should go in the `ImagePipeline` pytree. The`AbstractScatteringPotential` may need access to the wavelength. The wavelength is currently housed in the `CTF` class, however the CTF and scattering potential are unrelated.

(It is very important that we have no duplicate parameters in the `ImagePipeline` pytree. One way of testing this is with  [`equinox.tree_check`](https://docs.kidger.site/equinox/api/manipulation/#equinox.tree_check))

Whether or not this refactor is necessary for addressing #167, this will continue to come up, so wanted to mention now. For example, if we want to build out the model for the DQE in the `AbstractDetector`, it will also need the wavelength—as well as the dose rate. This is just one example. Many components of an image formation pipeline will specifically want to know about the wavelength and the dose rate.

It’s not completely clear how to address this in my mind, but something like the following skeleton is the best I have right now:

- I’d like the CTF to continue to not need required parameters at call time besides a coordinate system. Therefore, we can keep the same API with `CTF.voltage_in_kilovolts`. However, we can mark the parameter with `field(static=True)` so that it is not treated as a pytree leaf. Then, we can provide an option to pass the CTF a voltage at call time that overwrites the voltage stored in the CTF (this is sadly a bit hacky).
- Give the acceleration voltage to the `ElectronDose`. Now, similar to in `AbstractDetector.__call__`, we can pass the `ElectronDose` to `AbstractOptics.__call__`. Then, the voltage can just be passed to the CTF.
- Now, we can probably just keep passing around the `ElectronDose` at call-time to methods that need it. For example, the `ElectronDose` can be passed to methods of the `AbstractScatteringPotential` or the `AbstractSpecimen`. This is similar to how the `ImageConfig` is currently used. This will make it a required object in the pytree, so we can change the `Instrument.__init__` (currently it is optional). 

We can look into if using [`equinox.nn.Shared`](https://docs.kidger.site/equinox/api/nn/shared/#sharing-layers) might be appropriate for this task, but it seems to me the above is better.

This should also be general enough to handle future models of radiation damage, which will be housed in the `ElectronDose` :)

mjo22:
Rethinking this, I think it would be better to just always mark the `voltage_in_kilovolts` as static. I don't think we would ever want to have this as a trainable parameter.

mjo22:
[...]

Issue #167, opened 2024-02-27
Modeling the specimen phase shift distribution


mjo22:
Need to make sure that rasterized voxel grid scattering potentials have necessary factors out front so that their 2D projection images represent phase shifts in the exit plane. See function ReturnScatteringPotentialOfAVoxel here: https://github.com/timothygrant80/cisTEM/blob/master/src/core/scattering_potential.h

Also, there is a question in my mind of whether or not we should be using the error function to evaluate voxel grids at a given pixel. For reasons, see here: https://journals.iucr.org/m/issues/2021/06/00/rq5007/index.html

ehthiede:
I feel like there's multiple factors missing.  That said, it's a bit tricky to make sure we have them all without going into the math you are using for the detector model (I'm afraid I haven't been staying very caught up there).  Moreover I'm afraid the linked CisTEM code wasn't helpful for me (maybe I was looking in the wrong places).  Do you have a mathematical reference for your model you could share?  If it helps to have something to refer to, I've been looking at Chapter 4.3 of the _IUCr International Tables for Crystallography, Volume C,_ Chapter 4.3.

The error function is an interesting question.  I think in the end it comes down to how finely we need to parameterize our volumes (in real space) due to other factors, e.g. to avoid spurious correlations as a function of rotation.  I feel like we should get some numerical tests on the effect of resolution before we commit one way or the other.

mjo22:
The reference used for the cisTEM scattering potential seems to be the appendix of Vulovic et al 2013 (https://pubmed.ncbi.nlm.nih.gov/23711417/). I believe that the object we are after is $\sigma V_z(x, y, z)$, given in Eq. 22. We should carefully work this out...

jkawamura8:
Hello, Kawamura here.

geoffwoollard:
Regarding using the error function(s) instead a gaussian, I have looked into this numerically.

It's related to how small the pixel size is versus the decay length of the Gaussian. Think area-under-the-curve from integral calculus. 

The Gaussian drops off to exp(-x^2/2) when we are x standard deviations away. That's exp(-4.5)=1% at 3 std!

So if the pixel size is too large, then the Gaussian decays to near zero. Hence we need to integrate the Gaussian density inside the pixel. Hence the (difference of) error function(s). You get a difference in the x and y direction, so err_1_x - err_2_x and err_1_y - err_2_y, so 4 error function evaluations.

So either you have very tiny pixels and 
[...]

PR #200, opened 2024-03-28
Move the accelerating voltage from the `CTF` to the `Instrument`

mjo22:
Addressing #196.


mjo22:
Made the API changes! This should set us up to address #167. Settled on the following:

1. Move the voltage to the `Instrument`. This is an API change, where now it is required to pass a voltage to an instrument in order to initialize it. This also means that the `ImagePipeline` no longer treats an instrument as optional.
2. Pass the `wavelength_in_angstroms` to the `AbstractPotentialIntegrator`s. Now, we can convert the potentials to phase shifts and then integrate, or integrate and then convert to phase shifts--depending on the approximation.
3. Also cleaned up the `ImagePipeline` and the `Instrument` a bit.

mjo22:
Making this PR a place for minor general purpose fixes for the simulator internals as well. Just removed all `Null*` models from the library, like `NullOptics`. These are bad practice. Instead, just use `None`.

One example of this is that passing an image through an optics model changes what exactly the image represents. In the current paradigm, before an image is passed to the `AbstractDetector`, it is a squared wavefunction. After passing through the detector, it is a detector readout. A `NullDetector` returns the squared wavefunction unchanged, which is obviously not a detector readout.

mjo22:
Also removing `cryojax.typing`. This was a submodule used for array type hints. However, learning more about `jaxtyping` and run-time type checkers this was not a great idea.

mjo22:
@ehthiede what do we think about merging this and addressing #167 in a different PR? This PR should make these changes possible and future work in the atoms -> image pipeline easier.

After addressing #167, I would propose this as a marker for a new version release with the updated API.

Issue #201, opened 2024-04-01
Add a grid search into cryojax?


mjo22:
I am planning on writing a grid search so that I can do end-to-end parameter estimation in `jax`, and I am wondering if this has a place in `cryojax`. The plan is to write a very generic interface that can support different cost functions, algorithms, and ways of storing results.

A rough outline of the implementation I am imagining is the following:

1. *The grid:* The grid would be represented as an arbitrary pytree, whose leaves have leading "batch" dimensions. Namely, leaf $i$ is an array (or a pytree of arrays) with leading dimension(s) $N_i$. For $m$ leaves, the grid is then represented as a $N_1 \times N_2 \times ... \times N_m$ cartesian grid.
2. *The cost function:* The cost function $f$ would be a function that takes in a pytree of the same structure of the grid (and additional arguments), evaluated at grid point $(i_1, i_2, ..., i_m)$. This function can return a single value of the cost function, or it can return a *grid* of cost function evaluations. This would allow support for more clever grid searches than a simple exhaustive search, such as cross-correlation. In general, this sub-grid returned by $f$ would explore a region of parameter space unrelated to the $N_1 \times N_2 \times ... \times N_m$ grid.
3. *The solution:* There would need to be a flexible API for how the results of the grid search are stored. In a simple case, one could store the results of the best cost function evaluation at every grid point returned by the function $f$. In a more complicated case, one might want to use the grid search to marginalize away a portion of parameter space.

The question is: does this belong in `cryojax` or in a small separate library? In the meantime, will make a branch and work in `cryojax.inference`.

mjo22:
Will be using the `optimistix` API and internals for inspiration: https://github.com/patrick-kidger/optimistix

ehthiede:
Brings up an interesting point ... on one level we probably want some ability to optimize things out of the box, but too much could easily lead to feature-creep.  Maybe we can have an "optimize" entry-point with a default option or two, e.g. https://setuptools.pypa.io/en/latest/userguide/entry_point.html?  Grid search, or gradient descent could be the default option, which users could then overwrite using plugins... thoughts?

mjo22:
Agreed on the feature-creep point. I am a little confused on how the plug-ins would work, but if I am understanding correctly, I would even go a step further and say we should make even less assumptions about how the user will want to perform optimization. 

[...]

PR #207, opened 2024-04-10
Add a grid search API in cryojax.inference

mjo22:
Addresses #201. The API is very generic and assumes nothing about cryo-EM. It could be released to the JAX community, operating at a similar level of abstraction as something like `optax`. I think it will be really useful for everyone to have this as a part of a maintained package, but my dilemma is that I am a little hesitant right now to maintain something for a more general community outside of cryo-EM. So I'm not sure yet if its appropriate to add this into `cryojax` and keep it there, make it an independent package, or add it now and later make it an independent package.

Aside from this, this PR also includes important packaging updates. The submodules in `cryojax` were getting a little superfluous. For example, I've eliminated `cryojax.core` and absorbed `cryojax.io` into `cryojax.data`.


mjo22:
Whether or not this is added to `cryojax`, to finish this PR I would like to add an inference tutorial notebook into a documentation for a cross-correlation based grid search.

mjo22:
Finished the tutorial for the grid search and ready to merge. This PR now also includes the new API, so merging will mark a big version release.

PR #212, opened 2024-04-22
API Update in preparation for supporting more complex scattering theories. 

mjo22:
Needed to make the current code orthogonal to future updates. This is done through the new `AbstractScatteringTheory` interface. As a part of this, making a large-scale overhaul in a move toward a `cryojax` 1.0 API. When merged, this will mark a `0.3.0` version release.

Some highlights include

- As mentioned, there is a new `AbstractScatteringTheory` abstraction. For now, the important concrete instance is `LinearScatteringTheory`. Going forward, multislice can become a scattering theory.
- `ImageConfig` -> `InstrumentConfig`, and `Instrument` has been eliminated.
- `AbstractSpecimen` -> `AbstractStructuralEnsemble`, and `Specimen` -> `SingleStructureEnsemble`
- `AbstractPotentialIntegrator` -> `AbstractPotentialProjectionMethod`
- `AbstractOptics` -> `AbstractTransferTheory` and `CTF` -> `AberratedCTF`. `CTF` also picks up an abstract interface called `AbstractCTF`.
- Each concrete `AbstractPipeline` now simulates a distinct physical quantity, rather than it being ambiguous what the `ImagePipeline` simulates. Now, `ImagePipeline` has broken into `ContrastImagingPipeline`, `IntensityImagingPipeline`, and `ElectronCountsImagingPipeline`, which simulate the contrast, intensity (or squared wavefunction), and electron counts, respectively.

This fixed a lot of bad practices in the simulator internals. Before, it was unclear how we would go about using `cryojax` for wave scattering.


geoffwoollard:
Can you get the tests passing?

docs/index.md
```diff
@@ -42,66 +42,62 @@ The [`jax-finufft`](https://github.com/dfm/jax-finufft) package is an optional d
...
+# ... now, build the ensemble. In this case, the ensemble is just a single structure
+structural_ensemble = cxs.SingleStructureEnsemble(potential, pose)
 ```
 
-Next, build the model for the electron microscope. Here, we simply include a model for the CTF in the weak-phase approximation (linear image formation theory).
+Here, the 3D scattering potential array is read from `filename`. Then, the abstraction of the scattering potential is then loaded in fourier-space into a `FourierVoxelGridPotential`. The scattering potential can be generated with an external program, such as the [cisTEM](https://github.com/timothygrant80/cisTEM) simulate tool. Then, the representation of a biological specimen is instantiated, which also includes a pose and conformational heterogeneity. Here, the `SingleStructureEnsemble` class takes a pose but has no heterogeneity.
+
+Next, build the *scattering theory*. The simplest `scattering_theory` is the `LinearScatteringTheory`. This represents the usual image formation pipeline in cryo-EM, which forms images by computing projections of the potential and convolving the result with a contrast transfer function.
```
geoffwoollard, line 71: I would put some references here to the weak phase object approximation, and projection approximation.

src/cryojax/simulator/_transfer_theory/contrast_transfer_theory.py
```diff
@@ -1,27 +1,37 @@
...
+        frequency_grid_in_angstroms: Float[Array, "y_dim x_dim 2"],
+        *,
+        wavelength_in_angstroms: Optional[Float[Array, ""] | float] = None,
+        defocus_offset: Float[Array, ""] | float = 0.0,
+    ) -> Float[Array, "y_dim x_dim"] | Complex[Array, "y_dim x_dim"]:
+        raise NotImplementedError
 
 
-class CTF(AbstractFourierOperator, strict=True):
+class AberratedCTF(AbstractCTF, strict=True):
```
geoffwoollard, line 34: Is this the right name?

Perhaps sine CTF with constant amplitude contrast is the most descriptive and accurate.

geoffwoollard:
What will Gaussian additive noise be? In `ElectronCountsImagingPipeline`? That name may give the impression the outputs in integer counts. Some observed images will have integrated counts (e.g. detector in integrating mode) and not be integers.

mjo22:
Additive gaussian noise will be done through something from `cryojax.inference.distribution`! 

And interesting, I see. Currently the `PoissonDetector` will return integer counts, so I suppose it is accurate from the standpoint of the current state of the code. What do you think another name could be?

docs/index.md
mjo22: Maybe we should have a references section in the README?

src/cryojax/simulator/_transfer_theory/contrast_transfer_theory.py
mjo22: Yes wasn’t totally sure about this… definitely open to changes here.

PR #214, opened 2024-04-23
Settle on naming choices in new API

mjo22:
This PR is dedicated to settling on naming choices in the new API, addressing some comments by @geoffwoollard in #212.

Since the last PR, 

- `AberratedCTF` -> `ContrastTransferFunction`. If we ever model a more complicated CTF perhaps we can rethink this.
- Local variable names changes for `config` -> `instrument_config` and `pipeline` -> `imaging_pipeline`. This is to reflect base class names.


mjo22:
Cleaned up the code under-the-hood to allow for an implementation of imaging on the ewald sphere.

PR #216, opened 2024-05-06
API re-vamp of the atoms -> voxel grid pipeline.

mjo22:
This PR gives a new API and clean-ups to the atoms -> voxel grid pipeline. Basically, the old `from_atoms` method has now been removed and building voxel grids gets its own API. I think this makes the code quite a bit cleaner---I think I had an awkward recommendation a while ago in my thinking we should use the class constructor wrapping scheme.

More specifically, the PR includes

- Re-vamp the old `AtomCloud` into two objects `GaussianMixtureAtomicPotential` and `PengTabulatedAtomicPotential`. The first has parameters named generally for mixtures of gaussians, and the second has parameters named explicitly for parameters which are the results of fits to scattering factors (I think it’s good to make the distinction between these two things).
- These objects have a method `as_real_voxel_grid`.
- As a result, `from_atoms` constructor is gone. The input to this method was too cumbersome as I was wrapping the constructor `from_real_voxel_grid`. Also, I think it’s good to always make the user call a method that returns the voxel array. One reason is that it was unclear before how the user could do something as simple as pre-process the 3D voxel array before loading it into a `FourierVoxelGridPotentual`.
- Improve some naming clarity in form factor loading. In particular, `default_form_factor_params` -> `peng_element_scattering_factor_parameter_table` and `get_form_factor_params` -> `get_tabulated_scattering_factor_parameters`. Before, I was getting a little confused because a parameter table over all atoms was named something similar as the parameters for a specific atomic structure. Also, I'm seeing in the literature that form factors are the fourier transform of the density, and scattering factors and the fourier transform of the potential.

See the docstring of `PengTabulatedAtomicPotential` for how to load the object from tabulated scattering factors.


mjo22:
Just added the changes from #218. I also was able to work out the math myself and wrote up a draft of an explanation of it in the docstrings. If anyone has comments lmk!

PR #218, opened 2024-05-07
Added appropriate prefactors in _voxel_potential

jkawamura8:
This should properly adjust the Gaussians that we're approximating the neutral atoms as to correspond correctly to the scattering factors (and accompanying Peng paper conventions). I've also adjusted the test_voxels_from_atoms.py to add a corresponding test of the Fourier transform, to recover the equation for the scattering factors.


mjo22:
I've added the changes here into #216, this is great! Thanks! I've also written out the results of my working through the math in this PR. I was able to arrive at adding the factor of 4 pi and also verifying what we already had written up.

PR #225, opened 2024-05-17
implement gaussian mixture atomic potential integrator closes #224

### LINKED ISSUE #224: Implement integrator that simulates an image from an atom potential directly

DSilva27:
Implemented the class that integrates a GMM potential without converting it onto a volume first. Fixed some small issues in the `gemmi` parsers as well.


mjo22:
This looks great. There is a small bug I think though. You have the type hint for the `upsampling_factor` to be `float | int`, but you compute the upsampled shape assuming its an `int`. We can either support the case of `float` (which could actually be slightly annoying with the implementation of `downsample_with_fourier_cropping` function I have), or we could just support int?

mjo22:
Merging and re-opening in new branch!

PR #228, opened 2024-05-17
change type hint for upsampling_factor in GaussianMixtureProjection

DSilva27:


mjo22:
Added this in a commit to avoid going into a merge editor

PR #232, opened 2024-05-22
Add feature to compute images from extracting surfaces on the ewald sphere

mjo22:
Need to add testing, validate the code is right, and perhaps add references / math to the documentation.


mjo22:
Closes #250. Testing and validation to come!

PR #256, opened 2024-08-08
add option to only load starfile parameters in RelionDataset

DSilva27:


mjo22:
Great. I'll look through this and see if there are any touchups to add. One small thing is that calling a method `__mymethod__` is convention for a python magic method (a method in `object` that has special functionality), like `__getitem__`, `__mul__`, `__len__`, etc. I would rename the other methods to a private method convention, like `_mymethod`.

DSilva27:
Names changed.

mjo22:
Just pushed some small changes

mjo22 (approved):

DSilva27:
I can create a starfile with a couple of very small particles. Maybe 5 particles of size 8x8.

mjo22:
That would be awesome thank you... just added option to read JAX arrays on the CPU. This should avoid array copying when using the torch dataloader, but haven't confirmed this for sure.

mjo22:
We can go ahead and merge I think. Can continue to work on tests

PR #258, opened 2024-08-09
First pass at multislice implementation

mjo22:
This should be pretty much good to go. I think it would be good to formalize the multislice.ipynb into a more complete tutorial for the PR.

Fresnel propagator units look reasonable, amplitude contrast is implemented (needs references), and we have a naive implementation of variable slice thickness.

Anti-aliasing can be done with padding, but if there is another implementation that we should use instead we can do that.

The main issue is that things are slow because we rasterize a voxel-grid every time we want an image. Perhaps in the future an interpolation method should be used on a pre-rasterized grid.

Questions:
- Do we want to figure out how to add an envelope function to the `WaveTransferTheory`?


geoffwoollard:
Let me work on a tutorial for the PR

mjo22:
Two comments

- This branch will need some small updates to be in line with the recent updates in the `ContrastTransferFunction`.
- An idea for interpolating to a new voxel grid at a different pose: what if we used a 3D non-uniform FFT to do this? i.e. treat the rotated points as non-uniform points and compute a new grid.

mjo22:
So that things don't get out of date, I will go ahead and merge after fixing conflicts. Until it can be properly tested, multislice and other scattering approximations will stay in `cryojax.experimental`

PR #264, opened 2024-09-09
252 add envelope function to reliondataset

DSilva27:


mjo22:
After the "Particle Extraction" job in relion, the rlnCtfBfactor and rlnCtfScalefactor are unestimated and all set to 0 and 1, respectively. In a case like this, do we want to read in particle stacks with these fields or just set the envelope to `None`?

I guess an argument for always setting them is that we may not want behavior where depending on what exactly is in the STAR file, the fields change.

mjo22:
Yeah, I suppose a user can always just setup a new transfer theory without the envelope if they want

PR #272, opened 2024-09-20
Small bug in how the name of the mrcs was defined inside the starfiles

DSilva27:
The mrcs were not being named using a relative prefix, they were being named `index@full_path_to_relion_project/mrcs_name.mrcs`


mjo22:
approved

PR #306, opened 2025-01-13
Refactor transforms interface

mjo22:
Revamped and simplified the API for the `AbstractPyTreeTransform` in `cryojax.inference`. These classes are included in `cryojax` to make it easier for users to follow `equinox`'s recommendations for attaining custom behavior on a per-leaf basis. This is very similar and takes parts from the package `paramax`: https://github.com/danielward27/paramax


mjo22:
These "transforms" are not necessarily something for inference; they are a general utility for working with pytrees. Therefore, they are on the same level as `get_filter_spec` and belong in a new `cryojax.utils` module for JAX/Equinox utilities.

Issue #313, opened 2025-01-16
Implement imaging on the ewald sphere


mjo22:
In `cryojax.experimental`, there is a prototype for extracting surfaces from the ewald sphere. However, the CTF needs refactoring to implement the transfer theory for this functionality. Related to #285

mjo22:
Closed with #383

PR #314, opened 2025-01-16
Implement transfer theory for ewald sphere

mjo22:
Addresses #313


mjo22:
transfer theory not implemented yet, but merging API updates

PR #318, opened 2025-02-07
Implement error function evaluation in gaussians-to-atoms projections and debug projection upsampling

### LINKED ISSUE #284: `GaussianMixtureProjection` may not give correct results when using `upsampling_factor`
When I compute images with `upsampling_factor`, the overall contrast of my images greatly increases and I seem to reach some sort of instability at certain values. There may be a bug somehwere.

This is also making me wonder if we should move over to using error functions, like we do in the 3D case for `PengAtomicPotential.as_real_voxel_grid`. I'm seeing some good agreement between the error function and gaussian approach when using b factors, but not if b factors are zero.

mjo22:
Will close #284 and also implement new functionality to optionally evaluate error functions when taking projections.


src/cryojax/simulator/_potential_integrator/atom_potential_integrator.py
```diff
@@ -20,19 +23,38 @@ class GaussianMixtureProjection(
...
             This can be useful for reducing aliasing artifacts in the images.
+        - `use_error_functions`:
+            If `True`, use error functions to evaluate the projected potential at
+            a pixel to be the average value within the pixel using gaussian
+            integrals. If `False`, the potential at a pixel will simply be evaluated
+            as a gaussian.
+        - `atom_groups_in_series`:
+            The number of iterations used to evaluate the volume,
+            where the iteration is taken over groups of atoms.
+            This is useful if GPU memory is exhausted. By default, `1`.
```
DSilva27, line 53: This argument is a bit unclear. I would explain what a value of 1 means. For example "By default 1, which computes for all atoms at once".

src/cryojax/simulator/_potential_integrator/atom_potential_integrator.py
```diff
@@ -111,49 +127,183 @@ def compute_fourier_integrated_potential(
...
-    gauss_y = jnp.transpose(gauss_y, (2, 0, 1))
+    prefactor = 4 * jnp.pi * a[None, :, :] * b_inverse[None, :, :]
 
-    image = 4 * jnp.pi * jnp.sum(jnp.matmul(gauss_y, gauss_x), axis=0)
+    return prefactor * gauss_x, gauss_y
 
-    return image
+
+@eqx.filter_jit
+def _batched_map_with_contraction(fun, xs, batch_size):
```
DSilva27, line 290: This function might be unnecessary. I'll open an issue. I think this can be solved with jax.lax.map

DSilva27 (approved):

PR #320, opened 2025-02-19
fix small bug in batched_map functions

DSilva27:
Found that the way the batched map was implemented when going from [atoms to voxels](https://github.com/mjo22/cryojax/blob/124f957e1264f6935a7b085d2665f9c5aab4f6d1/src/cryojax/simulator/_potential_representation/atom_potential.py#L549) and when integrating an [atomic potential](https://github.com/mjo22/cryojax/blob/124f957e1264f6935a7b085d2665f9c5aab4f6d1/src/cryojax/simulator/_potential_integrator/atom_potential_integrator.py#L288) had a small bug. In some cases the function would not behave as expected, and the number of batches used was different.

For the second case, I simply changed the argument from batch_size to n_batches, as the function was never used with batch_size. For the first case, both instances were used, so I wrote a wrapper for the two cases.

### DIFF
--- a/docs/examples/compute-potential.ipynb
+++ b/docs/examples/compute-potential.ipynb
@@ -91,7 +91,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -101,7 +101,7 @@
     "real_voxel_grid = atom_potential.as_real_voxel_grid(\n",
     "    shape,\n",
     "    voxel_size,\n",
-    "    z_planes_in_parallel=10,\n",
+    "    batch_size_for_z_planes=10,\n",
     ")"
    ]
   },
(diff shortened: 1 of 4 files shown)


mjo22:
I'll add a quick name change to this PR as well!

mjo22 (approved):
This looks good! I bet there's still a way we could move to either the native `jax.lax.map` or the `cryojax.utils.batched_map` I recently added, but I guess if its not broken don't fix it!

DSilva27:
I don't think there is a way to make it work with `jax.lax.map`, it would require dividing the arrays to an array to be batched, and a remainder array. This is what we currently do. 

I have not checked `cryojax.utils.batched_map`. I'll try it out before merging the PR.

mjo22:
Ok, this makes sense. I think it will probably be the same with the `cryojax.utils.batched_map`.

Michael

PR #326, opened 2025-03-05
Adjust PDBReader

mjo22:
The `PDBReader` inspired from `mdtraj` seems to be using the static methods and class attributes from the mdtraj `PDBTrajectoryFile` object. We had also rewritten these static methods but were not using them; so I thought it made sense to remove them for simplicity. We could also rewrite things and not use the `PDBTrajectoryFile` entirely; but I think we should stick to one or the other.

However, there is an error for some reason I have not yet identified in the compute_potential notebook... the density looks weird and it is not centering for some reason. There is probably some small error somewhere


src/cryojax/io/_pdb.py
```diff
@@ -147,9 +185,246 @@ def read_atoms_from_pdb(
...
+                        formal_charge=atom.formal_charge,
+                    )
+                    atomByNumber[atom.serial_number] = newAtom
+
+    topology.create_standard_bonds()
+    topology.create_disulfide_bonds(atom_positions)
+
+    # Add bonds based on CONECT records.
+    connectBonds = []
+    for model in pdb.models:
```
DSilva27, line 397: There is a small bug here, which is also in my original implementation. Maybe these are not important for what we want, as we never go an error, but i think it's better to fix it

simply rewrite the for loop as 

```python
for l, model in enumerate(pdb.iter_models(use_all_models=True)):
    if self._i_model is not None and l != self._i_model:
        continue
```

DSilva27 (changes requested):

src/cryojax/io/_pdb.py
```diff
@@ -147,9 +185,246 @@ def read_atoms_from_pdb(
...
+        - `standard_names` : bool, default=True
+            If `True`, non-standard atomnames and residuenames are standardized to conform
+            with the current PDB format version. If set to false, this step is skipped.
+        - `topology` : mdtraj.core.Topology, default=None
+            If the `topology` is passed as input, it won't be parsed from the PDB file.
+            This saves time if you have to parse a big number of files.
+        """
+        # Check for errors
+        filename = pathlib.Path(filename)
+        _validate_pdb_file(filename)
```
DSilva27, line 246: These two lines break the url stuff, it should go inside the else

src/cryojax/io/_pdb.py
```diff
@@ -147,9 +185,248 @@ def read_atoms_from_pdb(
...
+            Should only be used if `is_assembly` is `True`.
+        - `standard_names` : bool, default=True
+            If `True`, non-standard atomnames and residuenames are standardized to conform
+            with the current PDB format version. If set to false, this step is skipped.
+        - `topology` : mdtraj.core.Topology, default=None
+            If the `topology` is passed as input, it won't be parsed from the PDB file.
+            This saves time if you have to parse a big number of files.
+        """
+        # Check for errors
+        filename = pathlib.Path(filename)
```
DSilva27, line 245: This one should also go after the url stuff. It checks if the path exists, a url is always going to raise an error.

DSilva27 (changes requested):

PR #330, opened 2025-03-07
321 physical solvent model

geoffwoollard:


docs/examples/solvent-models.ipynb
geoffwoollard, line 1: @mjo22 can you help figure out what is going on here with the normalization?

mjo22: Are you using the scale factors for the power directly from the Parkhurst paper? Or are you using scale factors in cryoJAX conventions?

src/cryojax/image/.DS_Store
DSilva27, line 1: What are these .DS_store? Is this something coming from your computer? Are they needed? Put them in the .gitignore

src/cryojax/simulator/_solvent.py
```diff
@@ -128,3 +131,139 @@ def sample_ice_spectrum(
...
+
+        if get_rfft:
+            return ice_spectrum_at_exit_plane
+        else:
+            return fftn(
+                irfftn(ice_spectrum_at_exit_plane, s=instrument_config.padded_shape)
+            )
+
+
+class Parkhurst2024_Gaussian(AbstractFourierOperator, strict=True):
```
DSilva27, line 210: Why is this defined twice? It also exists in `cryojax.image.operators`

tests/test_solvent.py
```diff
@@ -0,0 +1,73 @@
+# import jax
```
DSilva27, line 1: I don't think it's a good idea to have pieces of code that are just comments. Is there a plan to fix this before merging? Otherwise it should be removed from the PR.

DSilva27 (changes requested):

mjo22:
Hi, looking great so far! A few comments for getting this to the finish line:

1) Based on what you have done, here is a template for what I am thinking for the final PR (with summary + pseudocode)

- `power_envelope_function` -> `power_spectrum_function` because the function returns a 2D array
- Let's just include an option that changes how the solvent is sampled, since both are sampling gaussian random fields.
- We need to make sure we get the units right for implementation---the cryoJAX simulator computes images in absolute units. I see that `a_1` and `a_2` sum to 1, so I assume that they are just relative fractions of each gaussian. In order to compute in absolute units, we will need parameters for 1) An ice thickness parameter and 2) Energy per unit length ("Energy" in cryoJAX conventions which scale out fundamental constants; see here https://mjo22.github.io/cryojax/examples/compute-potential/). For 2), I think this must be related to their $V_0$ in the paper, which they compute from electron scattering factors of hydrogen and oxygen. We store these in `cryojax.constants`.
- I'm changing the API a bit for the PR; this code is left over from an early iteration of cryoJAX. I will need to update _scattering_theory.py for the below code to run. Changes are:
    - I think it may make sense to just change `Ice` -> `Solvent`. The name `Ice` is leftover from an early version of cryoJAX,      but local variable names are solvent and this seems to be the term people use most.
    - Rather than returning phase shifts at the exit plane, I think it would make more sense to return the integrated potential as this is what cryoJAX `potential_integrators` return. Therefore the multiplication by the interaction constant will get absorbed into the `WeakPhaseScatteringTheory` and names will change from `ice_phase_spectrum` -> `ice_integrated_potential`.

```python

"""
Abstraction of the ice in a cryo-EM image.
"""

from abc import abstractmethod
from typing_extensions import override

import jax.numpy as jnp
import jax.random as jr
from equinox import Module
from jaxtyping import Array, Complex, PRNGKeyArray

from ..image import fftn, irfftn
from ..image.operators import FourierGaussian, FourierOperatorLike
from ._instrument_config import InstrumentConfig
from ._scattering_theory import compute_object_phase_from_integrated_potential


class SolventMixturePower(AbstractFourierOperator, strict=True):
    r"""A model for the power of the gaussian random field (GRF) for solvent in cryo-EM.
    This implementation takes inspiration from Parkhurst et al. (2024). and uses its
    implementation as default.

    Parkhurst et al. (2024) models the power as a sum of two gaussians, with
    one gaussian modeling an envelope for the low-resolution decay of the power and
    another gaussian modeling the peak from the solvent at high-resolutions. This
    expression is given by

    .. math::
        P(k) = a_1 \exp(-k^2/(2 s_1^2)) + a_2 \exp(-(k-m_2)^2/(2 s_2^2)),

    where index `1` is for the envelope and index `2` is for the high-resolution peak.

    More generally, this class models the power as a mixture of any two functions, called
    `envelope_function` and `peak_function`.
    """

    envelope_function: FourierOperatorLike
    peak_function: FourierOperatorLike

    def __init__(
        self, 
        envelope_function: FourierOperatorLike | None = None,
        peak_function: FourierOperatorLike | None = None,
    ):
        if envelope_function is None:
            self.envelope_function = FourierGaussian(...)  # Set with default from Parkhust et al; need to take care to get conversion right between standard deviations and b factors
        if peak_function is None:
            self.peak_function = FourierGaussian(...)  # Again set with default from Parkhurst; need to implement a peak offset in the `FourierGaussian` class

    @override
    def __call__(self, frequency_grid: Float[Array, "y_dim x_dim 2"]) -> Float[Array, "y_dim x_dim"]:
        # I like this approach because its flexible and only makes core to cryoJAX the big ideas from Parkhurst
        # One computational downside right now 
        mixture_function = self.envelope_function * self.peak_function
        return mixture_function(frequency_grid)


class AbstractSolvent(Module, strict=True):
    """Base class for a model of the solvent in cryo-EM."""

    solvent_thickness_in_angstroms: eqx.AbstractVar[Float[Array, ""]]
    potential_scale: eqx.AbstractVar[Float[Array, ""]]

    @abstractmethod
    def sample_ice_integrated_potential(
        self,
        rng_key: PRNGKeyArray,
        instrument_config: InstrumentConfig,
        outputs_rfft: bool = True,
        outputs_real_space: bool = False,
    ) -> (
        Complex[
            Array,
            "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim//2+1}",
        ]
        | Complex[
            Array,
            "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim//2+1}",
        ]
    ):
        """Sample a stochastic realization of scattering potential due to the ice
        at the exit plane."""
        raise NotImplementedError

    def compute_integrated_potential_with_ice(
        self,
        rng_key: PRNGKeyArray,
        object_fourier_integrated_potential_at_exit_plane: (
            Complex[
                Array,
                "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim//2+1}",
            ]
            | Complex[
                Array,
                "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim}",
            ]
        ),
        instrument_config: InstrumentConfig,
        input_is_rfft: bool = True,
        outputs_real_space: bool = False,
    ) -> (
        Complex[
            Array,
            "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim//2+1}",
        ]
        | Complex[
            Array,
            "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim//2+1}",
        ]
    ):
        """Compute the combined spectrum of the ice and the specimen."""
        # Sample the realization of the phase due to the ice.
        ... # TODO: this function will also handle masking from a model for the solvent shell!
            # The shell will be a subclass of `cryojax.operators.AbstractRealOperator`
            # We can do this in a separate PR!


class GRFSolvent(AbstractIce, strict=True):
    r"""Ice modeled as a gaussian random field (GRF)."""

    solvent_thickness_in_angstroms: Float[Array, ""]
    potential_scale: float
    power_spectrum_function: FourierOperatorLike
    samples_power: bool

    def __init__(
        self,
        solvent_thickness_in_angstroms: Float[Array, ""] | float,
        potential_scale: float = ...,  # TODO: default value? use cryoJAX conventions for units of the potential, which is different from Parkhurst
        power_spectrum_function: FourierOperatorLike | None = None,
        samples_power: bool = False):
        """**Arguments:**

        - `thickness_in_angstroms`: 
            The solvent thickness in angstroms.
        - `potential_scale`:
            A dimensional factor that quantifies the characteristic scale
            of the potential. By default, this is calibrated from scattering factors in
            `cryojax.constants`.
        - `power_spectrum_function` :
            A function that computes the variance of the ice, modeled as colored gaussian noise.
            This function is treated as dimensionless, while `thickness_in_angstroms` and `potential_scale`
            determine the dimensions of the final image.
        - `samples_power`: 
            If `True`, the power spectrum of the sampled GRF is stochastic. If `False`,
            the power is deterministic (i.e. the GRF is generated by sampling uniform phases).
        """
        self.power_spectrum_function = power_spectrum_function or SolventMixturePower()
        self.samples_power = samples_power
        self.potential_scale = potential_scale
        self.solvent_thickness_in_angstroms = error_if_negative(solvent_thickness_in_angstroms)

    @override
    def sample_ice_integrated_potential(
        self,
        rng_key: PRNGKeyArray,
        instrument_config: InstrumentConfig,
        outputs_rfft: bool = True,
        outputs_real_space: bool = False,
    ) -> (
        Complex[
            Array,
            "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim//2+1}",
        ]
        | Complex[
            Array,
            "{instrument_config.padded_y_dim} {instrument_config.padded_x_dim}",
        ]
    ):
        """Sample a realization of the ice integrated potential as a gaussian random field."""
        ...
```

Also, some logistical comments:

- The final PR should be limited to adjustments to the ._solvent.py 2) The tests (the test_solvent.py looks great, thanks for writing this!) 3) Updated tutorial in simulate-image.ipynb. Make sure local changes, i.e. additionally committed files besides those in `cryojax/tests/data` and locally ran notebooks are kept the same. In the future PR, we can also start the docs for the solvent :)
- I apologize, I actually just updated the API for the solvent due to updates to the scattering theory. See the branch `354-booleans`, which will mark a `0.4.0` release.
- I actually can't checkout this branch on my mac because of the committed .DS_Store file. In the future, we can add this to the .gitignore

mjo22:
I thought it would be easiest to just push the refactoring and new API into PR #354. Can we integrate the work done in this branch into there? In particular: the test_solvent.py, waterbox in cryojax/tests/data, and work done in ._solvent.py? Let me know if there is anything I'm missing.

I also went ahead and created a skeleton for the solvent in the docs and for the Parkhurst 2024 solvent parameters in `cryojax.constants`. The _solvent.py should import constants from here to declare the default power.

Issue #332, opened 2025-03-27
pdb reader fails to import all atoms


shuyuvankerkwijk:
The new `read_atoms_from_pdb_or_cif` only registers roughly half of the atoms in pdb files.

I have documented this issue for two pdb files ([water_81_coords.pdb](https://zenodo.org/records/7862117) and relaxed_small_box_tip3p.pdb).  

1. Expected atom counts for water_81_coords.pdb:
- OH2: 16964
- Total: 50892 atoms

What `read_atoms_from_pdb_or_cif` loads:
- OH2: 9261 atoms (with filter 'not element H')
- Total: 27783 atoms

2. Expected atom counts for relaxed_small_box_tip3p.pdb:
- OW: 1471
- Total: 4404

What `read_atoms_from_pdb_or_cif` loads:
- OW: 887 atoms (with filter 'not element H')
- Total: 2672 atoms

For comparison, the legacy `read_atoms_from_pdb` function (tested using commit d1f638f) does not import hydrogens but otherwise successfully imports all expected oxygen atoms in both cases:
- Imports 16964 OH2 atoms from water_81_coords.pdb
- Imports 1471 OW atoms from relaxed_small_box_tip3p.pdb

Screenshots of the two different functions:
<img width="400" alt="Image" src="https://github.com/user-attachments/assets/e037be17-ee51-45ec-ab32-c785af451b37" />
<img width="400" alt="Image" src="https://github.com/user-attachments/assets/438bc49b-2a57-4367-a749-1c578eab2958" />

Maybe I am calling the functions wrong... but thought it worth bringing up!

mjo22:
There indeed seems to be a bug with the new reader. Will be making it a priority to look into this

mjo22:
Yes, we will simply drop the Biopython reader and fully move to mdtraj

DSilva27:
@shuyuvankerkwijk Could you checkout PR #333 (branch [332-pdb-reader-bug](https://github.com/mjo22/cryojax/tree/332-pdb-reader-bug)) and confirm that this is fixed for your systems? I checked one of them, but it'd be nice to have confirmation from you.

shuyuvankerkwijk:
@DSilva27 Just checked — your updated function imports the correct number of atoms for both .pdb files, with and without an atom filter!

mjo22:
Thank you for checking! Closed with #333

PR #333, opened 2025-03-27
Drop Biopython in favor of MDTraj for PDB reading

mjo22:
#332 points out that Biopython silently does not load all atoms if formatting isn't quite what the reader wants. Instead, move PDB reading to previous implementation by @DSilva27 in MDTraj.

As a result for now, mmCIF support will be dropped and support for assemblies will be dropped. When the below PR goes through in MDTraj, we will look again into mmCIF support

https://github.com/mdtraj/mdtraj/pull/2004


mjo22:
PR ready for review! Will close #332

DSilva27:
There were some tutorials using the old API, so I fixed them. I will wait for confirmation that everything is fine for the water models and then merge the PR.

PR #334, opened 2025-03-28
Give option to give per-particle arguments when generating a relion dataset

DSilva27:
THIS IS A DRAFT

Basically, I want the user to have the option to give a function that looks like

`def compute_image(relion_particle_parameters, static_args, dynamic_args): ...`

Such that for each image you would do: `compute_image(relion_particle_parameters[i], static_args, dynamic_args[i])`. 

In addition, I also modified the write starfile function so that static_args and dynamic_args are optional arguments. 

This would be extremely useful for computing heterogeneous datasets, as you have total control on how the configurations/conformations are used for the simulations (useful for defining custom distributions). Another case is simulating images with different noise parameters, or simply with any arguments that are not attributes of relion_particle_parameters (what Miro wanted to do).

I was thinking about having a has_aux option, but I think that covers that case as the user can just pre-compute whatever they need (although maybe this is something that we could consider for the future, not sure what a use case would be though).

Since this is a draft I was a bit lazy. Currently this assumes that the keywords for the arguments in the function given by the user are static_args and dynamic_args. That is, if the user defines these arguments by any other name the function fails. I did this to avoid having a bunch of if statements (since static_args and dynamic_args are optional now)

TODOs:
- Allow for general keywords for the static_args and dynamic_args function arguments. The position will be strict though. I am not too experienced with using `**kwargs`, but I assume that's what they are for?
- I did not adapt the docstring (besides the attributes), the examples need to be adapted.
- Better names for the arguments?


DSilva27:
Examples of usecase: images at different SNRs

```python

from typing import Any
import cryojax.simulator as cxs
from cryojax.inference import distributions as dist
... # other imports

def build_distribution_from_relion_particle_parameters(
    relion_particle_parameters: RelionParticleParameters,
    args: Any,
    variance: Float,
) -> dist.IndependentGaussianFourierModes:
    potential, potential_integrator = args

    structural_ensemble = cxs.SingleStructureEnsemble(
        potential, relion_particle_parameters.pose
    )

    scattering_theory = cxs.WeakPhaseScatteringTheory(
        structural_ensemble,
        potential_integrator,
        relion_particle_parameters.transfer_theory,
    )
    imaging_pipeline = cxs.ContrastImagingPipeline(
        relion_particle_parameters.instrument_config, scattering_theory
    )
    distribution = dist.IndependentGaussianFourierModes(
        imaging_pipeline,
        signal_scale_factor=jnp.sqrt(instrument_config.n_pixels),
        variance_function=op.Constant(variance),
    )
    return distribution

def compute_image_with_noise(
    relion_particle_parameters: RelionParticleParameters,
    static_args: Any,
    dynamic_args: Any,
):
    key, variance = dynamic_args
    distribution = build_distribution_from_relion_particle_parameters(
        relion_particle_parameters, static_args, variance
    )
    return distribution.sample(key)

seed = 0
key = jax.random.key(seed)
keys_noise = jax.random.split(key, number_of_images)
variances = jnp.linspace(
    start=0.5,  # start variance
    stop=2.5,   # end variance
    num=number_of_images
)


args = (potential, potential_integrator)  # the args compute_image takes
dynamic_args = (keys_noise, variances)

write_simulated_image_stack_from_starfile(
    metadata_for_noise_images,
    compute_image_with_noise,
    static_args=args,
    dynamic_args=dynamic_args,
    overwrite=True,
    is_jittable=True
)
```

DSilva27:
This should be ready for merge.

mjo22 (approved):
Looks great!

Issue #327, opened 2025-03-06
Remove `mdtraj` private method / attribute calls in the PDB reader


mjo22:
It would probably be best to get rid of these in the new PDB/mmCIF reader; they seem pretty straight forward to rewrite. Most of these calls are on static attributes / methods on the PDBTrajectoryFile class.

mjo22:
Closed with #335

PR #335, opened 2025-03-28
More cleanups to PDB reader

mjo22:
Address #327 and simply our PDB reader. Also, fix bug in standardizing residue and atom names.


DSilva27 (approved):
Tested with different PDBs, including a waterbox and it seems to work well.

PR #337, opened 2025-03-31
One particle in stack

Miro-Astore:
I noticed that when one tries to write a particle stack with just one image, there is a bug in the writer function. A simple if statement avoids an indexation error.


mjo22:
Just pushed updates; @Miro-Astore can you take a look and make sure things are ready to go?

mjo22 (approved):
Working well!

PR #338, opened 2025-03-31
set jittable default to false for writing image starfile

Miro-Astore:
Seems to me like makign the default as false for image writing is safer, prevents opaque errors for new users.


mjo22 (approved):
Looks good!

Issue #331, opened 2025-03-24
Broadcast opticsGroup parameters in RELION dataset


mjo22:
Currently, we load RELION opticsGroup parameters without a batch dimension, and particle metadata with a batch dimension in the `cryojax.data.RelionParticleMetadata` loader. This makes end-user code a little annoying because one needs to specify a `filter_spec` for the vmapped and non-vmapped args at the level of individual fields when reading multiple particles.

The point of RELION integration is to make things simple, so I think it makes sense to just broadcast all opticsGroup metadata. As a result, all arrays in `InstrumentConfig` and `ContrastTransferTheory` upon reading from RELION will have batch dimensions (i.e. the pixel size, the voltage, ...).

Therefore, `RelionParticleParameters` objects and all of its nodes can pass directly over vmap boundaries.

mjo22:
Closed with #341

PR #341, opened 2025-04-03
Updates to relion interoperability

mjo22:
Did some renaming for clarity of the various relion objects. Also, addressing #331 to give batch dimension to all parameters in the `RelionParticleParameters`. Need to still get the relion simulated dataset functionality working again


DSilva27:
updated starfile writing. @Miro-Astore I checked that you can generate datasets with 1 image.

DSilva27:
It is not completely true that all parameters have a batch dimension. There are things missing like "shape" or "pad_mode". Of course, it doesn't make sense to have a batch dimension for these, but it is still important, as vmapping is not direct. For vmapping you need to use

```python
def myfunction(pararameter_reader, args):
     return something

myfunction_vmapped = eqx.filter_vmap(myfunction, in_axes(eqx.if_array(0), None))
myfunction_vmapped(parameter_reader[0:n], args)
```

mjo22:
I think things are basically ready to merge!

PR #342, opened 2025-04-04
340 write an ensemble reweighting tutorial

DSilva27:
Tutorial for simulating data from structural ensembles (continuous heterogeneity). I also included a section for computing the likelihood matrix for an ensemble vs dataset. The tutorial includes the usage of external dataloaders.

I wanted to cover many things that I get asked constantly. Hopefully this will be helpful. I think Luke, Pilar and Minhuan need this as soon as possible.


DSilva27:
This branch was not created from main, but from  #341 as I did not want to have to update the tutorial as soon as we rolled in this PR.

aevans1:
I have a couple of comments here, still looking through! Thanks again for writing this, its a good exercise in learning all the constituent parts as well with cryojax.

- currently the [groel_chainA_holo.pdb](https://github.com/mjo22/cryojax/blob/340-write-an-ensemble-reweighting-tutorial/docs/examples/data/groel_chainA_holo.pdb) is empty, so throws an error in pdb reader about its atom indices
- it may be good to do the atomic potential by default rather than fourier slice, even for just these 2 pdb files I think this is really memory intensive
- I think that the `compute_likelihood_matrix_with_lax_map` may need some looking at, it takes over 5 minutes on gpu for 1,000 images and 2 structures, whereas the other methods take about 30 seconds, something might be amiss here, haven't checked what yet

DSilva27:
> * currently the [groel_chainA_holo.pdb](https://github.com/mjo22/cryojax/blob/340-write-an-ensemble-reweighting-tutorial/docs/examples/data/groel_chainA_holo.pdb) is empty, so throws an error in pdb reader about its atom indices

Whoops. Thanks. I always forget that these things need to be added to git lfs.

> e.g, I pick volume x_1, find population count N_1 according to weight w_1, simulate N_1 images, then switch to volume x_2 in memory and do the same, never having all volumes loaded in at the same time.

If memory is a worry, sure, this can be done. Just do it for one volume and then stack the distance matrices. This can be done even in the context of the tutorial. Think about the tutorial as the case for 2 volumes, you can add or remove as you need. So you could have one volume, compute the likelihood matrix (now a vector) and then iterate to the next volume. I don't even think the code needs to be updated (up to dealing with concatenating the matrices). 


> * I think that the `compute_likelihood_matrix_with_lax_map` may need some looking at, it takes over 5 minutes on gpu for 1,000 images and 2 structures, whereas the other methods take about 30 seconds, something might be amiss here, haven't checked what yet

Did you play with the batch size of the dataloader and `jax.lax.map` that runs on the image subset?  That is, did you change `batch_size_images`? If not, then it will take long because it needs to run many batches in series, while the vmap runs everything at once.

DSilva27:
I checked and the file should be in git lfs. Are you sure you pulled it correctly? @aevans1

aevans1:
Hi @DSilva27 , i did not pull it correctly, once I did that pdb was fine, I think that for some reason on a regular git pull, one of the pdbs loads, but the lfs pulled is needed to get both.

And I am going back with the data loader that seems fine as well wqith different batch sizes, thanks for pointing it out.

aevans1:
@DSilva27 a small thing, maybe reading it wrong, it looks like the defocus in the notebook is going from 1000 to 1500 as a range, did you mean 10000 to 15000? Asking because in previous could we've passed around its been the latter range.

DSilva27:
Introduced @aevans1 changes. Thanks again for your help!

PR #343, opened 2025-04-07
Can now generalize to new particle readers in the future

mjo22:


DSilva27 (commented):
The more I read things like `is_loading_metadata`, the more I dislike it.

I think we should change

`is_loading_metadata` to `load_metadata`
`is_loading_on_cpu` to `load_arrays_on_cpu`
`is_broadcasting_optics_group` to `broadcast_optic_groups`
`is_loading_envelope_function` to `load_envelope_function`

mjo22:
These alternate names sound fine, we can make the change. The reason was so that the user can basically understand what parts of the class they can ignore if they are only interested in understanding the basic functionality of the reader; all of the different boolean options are important but make the reader somewhat difficult to digest and can be mostly ignored. They sound slightly less on in a future abstraction where we could have 

```python
with RelionParticleParameterReader(…) as param_reader:
    param_reader.is_loading_on_cpu = True
    …
```

How about we make the merge, and then in the other branch we can make the final name change?

PR #350, opened 2025-04-08
rename some arguments and update read-dataset tutorial to show all options when loading parameters

DSilva27:
Though it would be a good addition to show all options in the tutorial (even if just the default ones).

The reason I want to change the argument names is that I think they are more clear if they resemble spoken English more. I am not a native speaker, so maybe I'm biased

### DIFF
--- a/docs/examples/read-dataset.ipynb
+++ b/docs/examples/read-dataset.ipynb
@@ -79,6 +79,10 @@
     "param_reader = RelionParticleParameterReader(\n",
     "    path_to_starfile=\"./data/ribosome_4ug0_particles.star\",\n",
     "    path_to_relion_project=\"./\",\n",
+    "    load_metadata=False,  # Set to True if you want to load the raw starfile dataframes\n",
+    "    load_arrays_on_cpu=False,  # Set to True to load jax arrays on CPU\n",
+    "    broadcast_optic_group=True,  # This is useful for vmapping\n",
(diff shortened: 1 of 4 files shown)


mjo22:
This looks mostly good; should be `broadcast_optics_group` though and I think `load_envelope_function`? I think if the latter is `False` then the envelope function is loaded as `None`, so it would be accurate to say that the envelope is not being loaded.

Let me think this over, and we should ask for more feedback. A reason for these names has larger context: in general I have been slowly moving cryojax booleans to `is_*` convention so that booleans don’t look accidentally like functions (i.e. “load_metadata” more looks like a function to me, not a boolean), but this can be hard to stick to because, as you’ve pointed out, the convention is pretty limiting / can be awkward. Need to decide if we want to stop moving booleans to this convention then.

DSilva27:
hmm what you say about "it sounds like a function" makes sense. It would probably be better if the syntax is consistent in all functions. Think it over, if you don't like it, then let's just ignore this PR and merge the other one. If anyone ever complains, then we can think about this again.

mjo22:
See the latest PR in the #341 branch. I think this is a good middle ground. The names are less clunky, and I even removed one of them (loading data on CPU can instead just be done directly by explicitly passing the JAX device).

mjo22:
Closing, addressed in #341

PR #360, opened 2025-04-11
Fsc frc

Miro-Astore:
Added support for calculating fourier shell or fourier ring correlations using existing infrastructure for calculating power spectrums. This seems to be working well. However, I don't believe the power spectrum API is jit compatible so this is not either. Is this intentional or should I raise an issue? 

Attached you'll find examples of comparing cryosparc derived half-maps. 
cryojax:
image

cryosparc: 
image

I've also attached a zip file with the notebook and files I used to make this PR if you'd like to check it yourselves.

### DIFF
--- a/src/cryojax/coordinates/__init__.py
+++ b/src/cryojax/coordinates/__init__.py
@@ -4,5 +4,6 @@ from ._make_coordinate_grids import (
     make_1d_frequency_grid as make_1d_frequency_grid,
     make_coordinate_grid as make_coordinate_grid,
     make_frequency_grid as make_frequency_grid,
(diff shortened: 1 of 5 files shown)


mjo22 (changes requested):
Looks great, and great test with csparc! We should change type hints for `minimum_frequency`, `maximum_frequency` and `pixel_size` just to float to make clear that it is not jittable unless this is true. We should also do this in the power spectrum function.

Also, as we discussed I think it would make sense to just pass the images either in real or fourier, not both.

mjo22:
Closed with #363

PR #363, opened 2025-04-16
Add fourier shell correlation in `cryojax.image`

mjo22:
Also refactors the power spectrum code a bit for a name change and to make sure that things are easily jittable if `pixel_size` is passed.

This is branching off of #360. Mostly made a new branch to avoid merge conflict, since that branch was from before the API updates the other day.


src/cryojax/image/_fourier_statistics.py
```diff
@@ -0,0 +1,280 @@
...
+        not measured in inverse angstroms, even if `radial_frequency_grid`
+        is in inverse angstroms.
+
+    **Returns:**
+
+    A tuple of the radially averaged power spectrum and the frequency bins
+    over which it is computed.
+    """
+    # Compute squared amplitudes
+    squared_fourier_amplitudes = (
```
DSilva27, line 74: Why not just `jnp.abs(fourier_image_or_volume) ** 2`? Probably the same when it crosses the jit boundary, but it should be more optimal to do abs for non-jit cases.

DSilva27: whoops. you are right.

src/cryojax/image/_fourier_statistics.py
```diff
@@ -0,0 +1,280 @@
...
+    # ... get minimum of array
+    threshold_crossing_index = jnp.amin(indices_at_0_to_1_flips) + 1
+    frequency_threshold = frequency_bins[threshold_crossing_index]
+
+    return correlation_curve, frequency_bins, frequency_threshold
+
+
+def _make_radial_frequency_bins(shape, minimum_frequency, maximum_frequency, pixel_size):
+    q_min, q_max = minimum_frequency, maximum_frequency
+    q_step = 1.0 / max(*shape)
```
DSilva27, line 278: This used to be `q_step = 1.0 / (pixel_size * max(*shape))`. Is this change intended?

src/cryojax/image/_fourier_statistics.py
```diff
@@ -0,0 +1,280 @@
...
+    *,
+    minimum_frequency: float = 0.0,
+    maximum_frequency: float = math.sqrt(2) / 2,
+) -> tuple[Float[Array, " n_bins"], Float[Array, " n_bins"], Float[Array, ""]]:
+    """Compute the fourier shell correlation for two voxel maps.
+
+    **Arguments:**
+
+    - `volume_1`:
+        A volume in real or fourier space. If it is in fourier space,
```
DSilva27, line 166: This says that it can take real volumes, but it never computes the FFT in the case of real volumes.

src/cryojax/image/_fourier_statistics.py
mjo22: Yeah; otherwise the pixel size needs to be marked static under JIT, which was causing some confusion. This is slightly confusing and was a trade off, we can make sure to document

src/cryojax/image/_fourier_statistics.py
mjo22: Oops missed this! I’ll fix

Miro-Astore:
I've gone through and I get similar curves and thresholds to cryosparc. This is ready for merge. Here are the results of two half maps from a reconstruction in cryosparc in cryojax.
image

Here is the results of refinement from cryojax. Focus on unmasked half-maps. 
image

PR #379, opened 2025-05-06
Implement tests for cryojax.data._relion

DSilva27:


DSilva27:
ToDo:

- Starfiles for Helical data
- Tests for abstract class
- Tests for cases that raise an error
- Test for starfile that leads to defaults + mrcs with only one image

mjo22 (approved):
Looks good!

PR #381, opened 2025-05-13
Move helical simulator out of package

mjo22:
This belongs downstream in my own work. This PR also includes changes to make `structural_ensemble` modeling more flexible downstream


DSilva27 (commented):
Are we removing the simulator but not the ability to read helical datasets?

DSilva27 (approved):

PR #382, opened 2025-05-16
Add soft-edge rectangular masks 

mjo22:
A few things

- Adds rectangular masks
- Removes filter and mask properties as parameters. It doesn't really make sense to store these as parameters since computation is only done on `__init__`.
- Adds new abstract class `AbstractBooleanMask`, which precomputes a boolean array for the region of signal. The `AbstractGaussianDistribution` computed this array on-the-fly, but there was no reason to not have this in precompute.
- More flexibility in applying filters and masks in the `AbstractGaussianDistribution`. In particular, now there's an option to normalize using the mask but not apply the mask.


DSilva27 (approved):
Looks good!

PR #388, opened 2025-06-09
cryojax.data refactor for powerful STAR file writing API

mjo22:
Notable changes:

- The `RelionParticleParameterFile` and the `RelionParticleStackDataset` each include modes 'r' and 'w' for reading and writing STAR files. Additionally, several new features in each mode! For example, in reading mode we now have the ability to select particles for the dataset, rather than including all entries of the STAR file (for example, filter based on class number, CTF resolution, etc).
- The function `simulate_particle_stack` generalizes the old simulation function to use any `AbstractParticleParameterDataset`
- No more `RelionParticleParameters` and `ParticleStack` equinox modules. These are simply replaced by typed dictionaries. A few reasons of this are that 1) This turns out to make extending an `AbstractParticleParameterFile` and `AbstractParticleStackDataset` more flexible and 2) These really need only be generic python containers, and updating dictionaries, such as in a gradient descent loop, is much simpler than updating an equinox module using `eqx.tree_at` 3) When updating parameters via `parameter_file[index] = ...`, a plain dictionary can be passed, such as `parameter_file[index] = dict(pose=pose)`. This is simple and allows each dict key to be optional. There are a few other advantages I won't list here.
- The `RelionHelicalParameterDataset` has been removed

To support backward compatibility, we can update the old STAR file writing functions to use the new framework under the hood. We can then add a deprecation warning. 

Still need to continue to add tests and update tutorials (that is, if we don't go the route of using deprecation warnings). Some things are fully untested for now, such as the new `simulate_particle_stack` function (a generalized version of the old dataset simulation function) and the `selection_filter` in the `RelionParticleParameterFile` (previously called the `RelionParticleParameterDataset`).


src/cryojax/data/_particle_data/relion.py
```diff
@@ -0,0 +1,1597 @@
...
+            transfer_theory=transfer_theory,
+            metadata=metadata,
+        )
+
+    @override
+    def __len__(self) -> int:
+        return len(self.starfile_data["particles"])
+
+    @override
+    def __setitem__(
```
DSilva27, line 294: Assuming the current behavior is the desired behavior:

It should be clear that `__setitem__` should only be used to update parameters. In addition, it should raise an error when indices that do not exist are "set". For example, I generated a new parameters file, then tried `param_file[0:10] = parameters`, and it ran without issue, but it did not really update the parameters in the file, the file was still empty. This can be solved by using append, but it should either work with setitem or it should raise an error.

src/cryojax/data/_particle_data/relion.py
```diff
@@ -0,0 +1,1597 @@
...
+        )
+        particle_data = (
+            pd.concat([particle_data, particle_data_to_append], ignore_index=True)
+            if len(particle_data) > 0
+            else particle_data_to_append
+        )
+        self._starfile_data = StarfileData(particles=particle_data, optics=optics_data)
+
+    @override
+    def save(
```
DSilva27, line 360: The class has an overwrite option at initialization. This should be consistent with that option. Otherwise, one of them should be removed.

src/cryojax/data/_particle_data/relion.py
```diff
@@ -0,0 +1,1597 @@
...
+        try:
+            write_image_stack_to_mrc(
+                images,
+                pixel_size,
+                path_to_filename,
+                overwrite=self.filename_settings["overwrite"],
+                compression=self.filename_settings["compression"],
+            )
+        except Exception as err:
+            raise IOError(
```
DSilva27, line 715: Why is the "overwrite" option in filename settings? This does not really match with the initialization of the class. It also makes it so that overwriting is not possible.

DSilva27 (changes requested):

src/cryojax/data/_particle_data/relion.py
mjo22: This should raise an error—in this case the user has to use append, since these indices don’t yet exist (similar to list / array behavior). The _validate_dataset_index function should raise the error, we can see why that’s not doing so

DSilva27: Solved!

DSilva27:
We now have coverage for most functions. We are just missing the tests for triggering errors. Afterwards we should be all good to go!

DSilva27:
Missing test for appending for stack dataset.

PR #400, opened 2025-06-26
Move toward reaching stable definition of submodules

mjo22:
This is not ready for merge, but ready for discussion. The next stable release of cryoJAX should have submodules with agreed upon purposes and names.

Here, the proposed changes are

- `cryojax.utils` -> `cryojax.jax_util`
- `cryojax.inference` removed, for now this just becomes `cryojax.distributions`, but subject for discussion! the grid search tool goes into `cryojax.jax_util`, which is a much better fit I think.
- `cryojax.image.operators` -> `cryojax.image.operators` + `cryojax.image.transformers`; the former is for things that take in parameters and coordinates at run-time, the latter is for things that take in an image at run-time (i.e. filters and masks)


mjo22:
Closing for now. Most open questions are solved, besides whether or not cryojax.distributions should just be in cryojax.simulator. Punting this for a future PR.

PR #402, opened 2025-07-02
API simplifications

michael-0brien:
This PR is for simplifying the API. There are three changes here:

- The `AbstractStructualEnsemble` class is confusing because it is used for heterogeneity and also not. This instead becomes a subclass of `AbstractBiologicalStructure` and the `SingleStructureEnsemble` becomes `Structure` and local variable names becomes `structure`. Names are definitely subject for discussion!
- The new `AbstractBiologicalSpecimen` doesn't fully make sense as a part of the scattering theory class, rather, this should be defined on the `AbstractImageModel` and passed to the scattering theory class. Structures are distinctly modeled from how they are imaged.
- The `AbstractScatteringTheory` class is not necessary for most users. Address this fact by creating a `LinearImageModel`, which does not use the scattering theory.

### DIFF
--- a/src/cryojax/simulator/_potential_integrator/base_potential_integrator.py
+++ b/src/cryojax/simulator/_potential_integrator/base_potential_integrator.py
@@ -47,7 +47,7 @@ class AbstractVoxelPotentialIntegrator(
 ):
     """Base class for a method of integrating a voxel-based potential."""
 
-    pixel_rescaling_method: AbstractVar[Optional[str]]
+    pixel_rescaling_mode: AbstractVar[Optional[str]]
 
     def _postprocess_in_plane_potential(
         self,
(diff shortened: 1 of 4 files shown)


src/cryojax/simulator/_image_model/base_image_model.py
```diff
@@ -0,0 +1,283 @@
...
+    signal_region: eqx.AbstractVar[Optional[Bool[Array, "_ _"]]]
+
+    @abstractmethod
+    def compute_fourier_image(
+        self, rng_key: Optional[PRNGKeyArray] = None
+    ) -> ImageArray | PaddedImageArray:
+        """Render an image without postprocessing."""
+        raise NotImplementedError
+
+    def render(
```
DSilva27, line 61: Are we keeping this for backward compatibility? Could be worth it to add a deprecation notice

DSilva27 (changes requested):
We should use this opportunity to add the typed Dics used in the data module to the init. Would make typehinting much much easier.

Issue #398, opened 2025-06-24
Support reading/writing STAR files with symmetry expansions


mjo22:
When a protein has symmetry, it is common to symmetrize a reconstruction by expanding the number of “particles” in the dataset by writing the pose parameters of each asymmetric subunit for the same image.

It would be best that this is possible for users downstream, not for us to explicitly build anything into the software about symmetrization. To support this downstream, we would need to have support for writing particles where, say, the pose has a leading dimension for the index of the particle (in the current definition) and the trailing dimension is the number of asymmetric subunits.

I am not sure if this extension should be core functionality in existing objects, or should somehow be incorporated in a new object. I think the former may be easier.

Here is an example publication. where this is done: https://www.sciencedirect.com/science/article/abs/pii/S0022283607007590.

I think this is a standard practice though in different softwares!

mjo22:
Closed with #403

PR #403, opened 2025-07-11
Improvements for custom STAR file behavior

mjo22:
Small adjustments to address #398. This doesn't support this explicitly, but rather makes it possible to read the data from another STAR file and then use that to create a new one that is symmetry expanded. This required improving the ability to operate on custom columns in the dataframe for when `loads_metadata = True`.


mjo22:
Closes #398

Issue #428, opened 2025-09-18
Tutorial for simulating a particle stack not running


mjo22:
Tutorial in `simulate-relion-dataset.ipynb` not running. The bug seems to be when the CTF is applied:

```
...

17 defocus = defocus_in_angstroms + 0.5 * astigmatism_in_angstroms * jnp.cos(
     18     2.0 * (azimuth - astigmatism_angle)
     19 )
     20 defocus_phase_shifts = -0.5 * defocus * wavelength_in_angstroms * k_sqr
     21 aberration_phase_shifts = (
---> [22](/path/to/cryojax/src/cryojax/simulator/_transfer_theory/common_functions.py:22)     0.25
     23     * spherical_aberration_in_angstroms
     24     * (wavelength_in_angstroms**3)
     25     * (k_sqr**2)
     26 )
     27 phase_shifts = (2 * jnp.pi) * (defocus_phase_shifts + aberration_phase_shifts)
     29 return phase_shifts

...

ValueError: Incompatible shapes for broadcasting: shapes=[(0,), (128, 65)]
```

Output of `jax.print_environment_info()`:

```
jax:    0.6.2
jaxlib: 0.6.2
numpy:  2.3.2
python: 3.11.4 (main, Jun 15 2023, 07:55:38) [Clang 14.0.3 (clang-1403.0.22.14.1)]
device info: cpu-1, 1 local devices"
process_count: 1
```

Issue #426, opened 2025-09-18
Tests for filters and masks


michael-0brien:

mjo22:
Closed with #431

PR #429, opened 2025-09-18
cryoJAX 0.5.0 release

michael-0brien:
For now, the road to 0.5.0 is addressing #426 and #428.


src/cryojax/simulator/_volume_parametrisation/ensemble/discrete_ensemble.py
```diff
@@ -0,0 +1,77 @@
...
+from jaxtyping import Array, Int, PRNGKeyArray
+
+from ....jax_util import error_if_negative
+from ..base_parametrisation import (
+    AbstractVolumeRepresentation,
+)
+from .base_ensemble import AbstractConformationalEnsemble
+
+
+class DiscreteConformationalEnsemble(
```
DSilva27, line 15: I like the new sample method, but I think there should also be a selection method. That is, a way to redefine conformation.

michael-0brien: Addressed by #430

DSilva27 (approved):
Looks good!

docs/api/constants/conventions.md
geoffwoollard, line 1: is this a placeholder? it's looks very empty...

docs/api/constants/scattering_factor_parameters.md
geoffwoollard, line 1: is this a placeholder? it's looks very empty...

geoffwoollard (changes requested):
Most of the notebooks worked (with JAX 0.7.2 as you asked, and Equinox 0.13.1). However, some of the dev/old notebooks are broken, and should be updated. Specifically

```
docs/examples/dev/ewald-sphere.ipynb: ImportError: cannot import name 'read_peng_element_scattering_factor_parameter_table' from 'cryojax.constants' (/mnt/home/gwoollard/repos/cryojax/src/cryojax/constants/__init__.py)
docs/examples/dev/jax2torch.ipynb: FileNotFoundError: [Errno 2] No such file or directory: './data/groel_5w0s_scattering_potential.mrc'
docs/examples/dev/multislice.ipynb: ImportError: cannot import name 'read_peng_element_scattering_factor_parameter_table' from 'cryojax.constants' (/mnt/home/gwoollard/repos/cryojax/src/cryojax/constants/__init__.py)
docs/examples/dev/simulating-and-reweighting-ensembles.ipynb: ModuleNotFoundError: No module named 'cryojax.image.transform'
docs/examples/old/cross-correlation-search.ipynb: ImportError: cannot import name 'read_array_with_spacing_from_mrc' from 'cryojax.io' (/mnt/home/gwoollard/repos/cryojax/src/cryojax/io/__init__.py)
docs/examples/old/simulate-micrograph.ipynb: ImportError: cannot import name 'read_array_with_spacing_from_mrc' from 'cryojax.io' (/mnt/home/gwoollard/repos/cryojax/src/cryojax/io/__init__.py)
```

Other than that, all the links work in the main readmes and contributor guide, and docs are clear. 

I was unclear about the status and purpose of the docs/*.md files. Is this needed to build the documentation at https://michael-0brien.github.io/cryojax/? Or is it just empty and waiting to be built out?

docs/examples/old/jax2torch.ipynb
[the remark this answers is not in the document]
michael-0brien: git LFS pull is necessary, see CONTRIBUTING.md

michael-0brien:
Actually I was hesitant to test more python versions as this seemed excessive for every PR, but I figured out how to test more versions only when merging into main. I tested the min and max supported versions, and it looks like python 3.13 is failing. Since we have many dependencies I think it’s best to support at least one less version than the most up to date python, so let’s see if 3.12 passes.

.github/workflows/testing.yml
```diff
@@ -1,46 +1,46 @@
...
 
 jobs:
-  build:
-
+  test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.11"]
+        python-version: ${{ fromJSON(github.base_ref == 'main' && '["3.10", "3.12"]' || '["3.11"]') }}
```
geoffwoollard, line 15: Only 3.10 and 3.12 are testing. You still want to test 3.11, right?

michael-0brien: We can add 3.11, I did min and max versions for simplicity

michael-0brien:
Adding testing matrix for min/max supported python versions and both ubuntu/macOS. Python 3.13 seems to work fine, so let's stick with that instead of 3.12.

michael-0brien:
Also using uv for testing workflow as this cuts overhead significantly compared to native pip. Many repos seem to be moving over to uv for local environment installation.

Also getting rid of codecov PR comments as I find this distracting (it seems to be commenting on every PR, even if no changes were made to tests). Down the road, if there's interest we could configure more minimal settings in the codecov.yml. In the meantime, we can go to the badge URL for reports.

geoffwoollard:
If you are seeing error like: 

```python
cryojax/__init__.py:12: in <module>
    from .cryojax_version import __version__ as __version__
E   ModuleNotFoundError: No module named 'cryojax.cryojax_version'
```

then be sure to reinstall from the .toml file, since the source code directory was changed. Just run `pip install -e .` again, if you already have an install from the previous version.

docs/examples/structure-refinement.ipynb
DSilva27, line 1: This tutorial has some outputs like `An NVIDIA GPU...` that shouldn't be there. Either the outputs need to be removed, or it should be executed in a clean environment.

PR #430, opened 2025-09-19
Remove `DiscreteStructuralEnsemble` and instead add to tutorial

michael-0brien:
It's not clear how useful this class is in practice. Mostly, it's useful because the `jax.lax.switch` statement has a learning curve to figure out. This PR asks if it may be best to just add this to the tutorial where a particle stack is simulated.

The previous tutorial to this one discusses how to create extensions to cryoJAX. Therefore, this tutorial can build on this and show how to create the `AbstractVolumeParametrisation` class.

This is open for discussion, @DSilva27 and I discussed a few options of what to do with this class and this was one of several. Can describe the rationale more to others interested.


michael-0brien:
Merging, this option is easier for maintainers and better communicates how to use the repo in practice.

PR #489, opened 2025-11-20
Absorb submodules `coordinates`, `operators`, and `transforms` into `cryojax.ndimage`

### LINKED ISSUE #485: Absorb image-related submodules into `cryojax.ndimage`
I think that the average `cryojax` user has to interact with too many submodules. One issue is that to create a volume, a user has to interact with `cryojax.io`, `cryojax.constants`, and even at times `cryojax.atom_util`. This would be addressed by #481.

Another issue is that functions for operating on images is scattered: `cryojax.coordinates` is for pixel/voxel coordinate systems, `cryojax.ndimage` is for utility functions that expect pixels/voxels as input, `cryojax.ndimage.transforms` / `cryojax.ndimage.operators` is for callables that accept images / coordinates as input (filters, masks, and parametrized functions). I would suggest we absorb all of these into the `cryojax.ndimage` namespace, keeping the other aliases for deprecation. This will be easier to use and document.

`cryojax.coordinates` used to be separate because it was not necessarily going to be for pixels/voxels, but with the creation of `cryojax.atom_util` any atom-based coordinate operations can go in there.

### LINKED ISSUE #488: Put `AbstractVolumeRenderFn`s and `AbstractVolumeIntegrator`s in the same file as their `AbstractVolumeRepresentation`
For those who read through the codebase, I think it's confusing to have these three  in different files / folders. With recent changes, it's possible to put the source code for each of these in same place!

michael-0brien:
`cryojax.coordinates`, `cryojax.ndimage.operators`, and `cryojax.ndimage.transforms` are deprecated and absorbed into `cryojax.ndimage`. Addressed #485. Support will be dropped in v0.6.0.

This is to improve user experience; users can now run `import cryojax.ndimage as im` to get all pixel/voxel related functions in one submodule. Recent updates to the code have made this simplification possible.


michael-0brien:
Adding a fix for #488 here too as this need not really be in the changelog

PR #496, opened 2025-12-02
Add `load_tabulated_volume`, a utility for loading atomistic volumes 

### LINKED ISSUE #481: Utility function for instantiating volumes
Volume instantiation can be cumbersome because we require that functions on classes in `cryojax.simulator` (i.e. `__init__` and `from_*` constructors) work across JIT and vmap boundaries. Following the spirit of `make_image_model`, we should create a wrapper to make it easy to instantiate volumes. 

This could be two functions: `make_atom_volume` and `make_voxel_volume`.

michael-0brien:


cryojax/simulator/_api_utils.py
```diff
@@ -240,47 +269,181 @@ def make_image_model(
...
+def load_tabulated_volume(
+    path_or_mmdf: str | pathlib.Path | pd.DataFrame,
+    *,
+    output_type: type[
+        IndependentAtomVolume | GaussianMixtureVolume
+    ] = IndependentAtomVolume,
+    tabulation: Literal["peng"] = "peng",
+    include_b_factors: bool = False,
+    b_factor_fn: Callable[[NDArrayLike, NDArrayLike], NDArrayLike] = identity_fn,
+    selection_string: str = "all",
```
DSilva27, line 335: I think that `center` should be given the same treatment as `selection_string`. It's the only other parameter that I think is common enough that it would be annoying to have to write `pdb_options = {"center": True}` every time the function is used.

michael-0brien: Sure, we can do that. But the thinking was that True is the default, so this would only be necessary when center is False

DSilva27: Huh, I checked before commenting this and I saw a `False`, but you are right default is `True`. In that case I think it's fine.

DSilva27 (approved):

PR #502, opened 2025-12-05
More efficient scattering factor parameter loading

michael-0brien:
Improved backend for loading scattering factor parameters, and some corresponding changes to `cryojax.constants`. Typical interface to `cryojax.constants` is unchanged, but other functions that would be atypical to use are changed.


DSilva27 (approved):
Looks good.

Issue #358, opened 2025-04-10
Advanced scattering theory considerations


michael-0brien:
It would be cool to one day support more advanced scattering theory considerations. Three things that come to mind are: 1) more flexible modeling of amplitude contrast and 2) scattering approximations beyond the first born approximation 3) more accurate parameterizations of atomic potentials, such as from Lobato, I., and D. Van Dyck (2014).

Point 1) is particularly important, as this relates to things like inelastic scattering and therefore radiation damage.

We now basically have an API flexible enough where these things can be built (related to #285 and #355). Of course, new features shouldn't make the API too confusing to use.

mjo22:
Related to #416.

michael-0brien:
Closing as I think this is too abstract at the moment

PR #503, opened 2025-12-05
Add Lobato electrostatic potential parametrization

michael-0brien:
This was easy to do in the current framework, as the `IndependentAtomVolume` is built for arbitrary scattering factors. Shouldn’t make much of a difference compared to Peng, but good to have in the package. Related to #358.


docs/api/constants.md
```diff
@@ -23,6 +24,40 @@ Modeling the electron scattering amplitudes of individual atoms is an important
...
+        - 15: Phosphorus
+        - 16: Sulfur
+        - 17: Chlorine
+        - 19: Potassium
+        - 20: Calcium
+        - 25: Manganese
+        - 26: Iron
+        - 27: Colbalt
+        - 29: Copper
+        - 30: Zinc
```
geoffwoollard, line 49: good job on those metals! Although the values are for neutral atoms, and they would typically be ionized in biomolecules

geoffwoollard (approved):

geoffwoollard:
@michael-0brien I think some nice to haves are

- projection (although note that the bessel function are not *integrated* inside the pixel (like gaussians to erf)
Screen Shot 2025-12-06 at 10 28 06 AM
Screen Shot 2025-12-06 at 10 28 42 AM

Also, it was unclear if you had a test that matches the potential over a distance range (like in Fig 5b). I think this would be a good test to have high accuracy (to show how much they agree). The figure from the paper is from Silver, which was the element where they disagreed the most.

Screen Shot 2025-12-06 at 10 27 17 AM

cryojax/simulator/_volume/independent_atom_volume.py
```diff
@@ -70,6 +70,37 @@ def __call__(
         return jnp.sum(jax.vmap(gaussian_fn)(self.a, self.b), axis=0)
 
 
+class LobatoScatteringFactor(AbstractFourierOperator, strict=True):
```
geoffwoollard, line 73: So can you use this functional form with a fittable parameters? E.g. just use one hydrogenic atom and fit its a_1, b_1?

geoffwoollard:
Regarding the modified bessel functions of second kind, they are in scipy and are accuracy, but not currently in jax.scipy.special. But they will likely be down the road, since the modified first kind ones are
https://github.com/jax-ml/jax/issues/12402

docs/api/constants.md
michael-0brien: Good point. I suppose this way at least PDBs with metals won’t error out and can be read

cryojax/simulator/_volume/independent_atom_volume.py
michael-0brien: Yes! The `IndependentAtomVolume` accepts arbitrary scattering factors expressed as any `cryojax.ndimage.AbstractFourierOperator`. We could even modify the `LobatoScatteringFactor.__init__` to accept `a` and `b` without a leading dimension.

michael-0brien:
> @michael-0brien I think some nice to haves are
> 
> * projection (although note that the bessel function are not _integrated_ inside the pixel (like gaussians to erf)

So we do have a numerical implementation of this via `FFTAtomProjection` + `IndepedentAtomVolume`. This even gives integration within pixels via with sinc convolution (which the caveat is for this to be numerically accurate, the scattering factor in the fourier domain should decay sufficiently to zero). I think sticking to Fourier space is probably the way to go because projections are trivial (just take $q_z = 0$), and evaluating special functions can be expensive. The main advantage to real space in my mind is analytic rather than numerical anti-aliasing (e.g. with erf). It has the big disadvantage of scaling like the number of atoms to some power, rather than linear scaling! 

> Also, it was unclear if you had a test that matches the potential over a distance range (like in Fig 5b). I think this would be a good test to have high accuracy (to show how much they agree). The figure from the paper is from Silver, which was the element where they disagreed the most.

This would be good to do—I implemented a test that compares peng and lobato scattering factors for hydrogen only, since this is when they most closely agree. It was difficult to demand close agreement for more elements in fourier space. A real space test for more elements at longer length scales would probably be more successful!

geoffwoollard:
I looked more into the projection. There is a singularity at zero, which only goes away when you integrate. Unfortunately it doesn't go away K_0(aR) + RK_1(aR) as R -> 0. But at last for the scipy version there is numerical stability in the function for very close to the origin (aR like 1e-20; even at float32 instead of float64)

There are analytic forms of the integral for a *disk* pixel, but not for a square pixel. For that, one could upsample, and then convolve with a specific kernel.

Anyways, seems we'll have to wait for the right bessel functions in JAX. Even when they are ported, it would be good to check numerical stability, including of gradients. I've done that before and people go into it in the github issue above.

geoffwoollard:
> I think sticking to Fourier space is probably the way to go 

agreed

michael-0brien:
> I looked more into the projection. There is a singularity at zero, which only goes away when you integrate. Unfortunately it doesn't go away K_0(aR) + RK_1(aR) as R -> 0. But at last for the scipy version there is numerical stability in the function for very close to the origin (aR like 1e-20; even at float32 instead of float64)
> 
> There are analytic forms of the integral for a _disk_ pixel, but not for a square pixel. For that, one could upsample, and then convolve with a specific kernel.
> 
> Anyways, seems we'll have to wait for the right bessel functions in JAX. Even when they are ported, it would be good to check numerical stability, including of gradients. I've done that before and people go into it in the github issue above.

This is interesting, thanks for looking into this. I suppose this gets into Peng vs Lobato as well, and what role these formalisms have in modeling electrostatic potentials in cryo-EM. If Lobato’s role is to just instantiate a plausible potential, maybe the current implementation is enough. But if the functional form should have another role, like how gaussians are also used for coarse-grain modeling, maybe we need to think harder about this! Not really sure what that looks like.

geoffwoollard:
Don't forget about this equation for the real spaced 3D volume from the appendix. Not sure if it's useful...
Screen Shot 2025-12-06 at 12 29 19 PM

Issue #484, opened 2025-11-14
Backprojection and linear operators via `lineax`


michael-0brien:
In `cryojax.simulator`, I propose adding a function `make_linear_operator` that returns a `lineax.FunctionLinearOperator`. This would wrap an image simulator such that:

- Simulating images is expressed with matrix-vector abstraction, `image = operator.mv(volume)`
- “Backprojection” operators are obtained simply as `operator.T` (uses `jax.linear_transpose` under the hood). 

I’m thinking the following for function signature: 

```python
from collections.abc import Callable

import lineax as lx
from jaxtyping import PyTree

def make_linear_operator(simulate_fn: Callable, pytree: PyTree, where_in_structure: Callable) -> tuple[lx.FunctionLinearOperator, PyTree]:
    …
    return operator, vector  # vector is pytree leaves in the input space 
```

- `simulate_fn` is a callable that simulates an image. Accepts input `pytree` at runtime (such as an `image_model`)
- `where_in_structure` is used to separate `pytree` as `vector, args = eqx.partition(pytree, …)`

If a simulator is not a linear operator, `operator.T` should throw an error when called. I’m not sure if it’s worth the extra engineering burden to add the infrastructure for runtime checks; maybe best to just add to the docs a warning about this.

michael-0brien:
We could also have two functions `make_linear_operator` and `unpack_linear_operator`, so the user can convert back to the tree structure they had originally.

PR #504, opened 2025-12-05
Add `make_linear_operator`: simulate images using `lineax`

michael-0brien:
We can simulate images using the `lineax.FunctionLinearOperator` for functions that are linear maps from volumes to images. This addition converts cryoJAX simulators to the `lineax` abstraction and therefore makes it very easy to access backprojection operators for building data analysis downstream. Addresses #484.

### DIFF
--- a/cryojax/experimental/__init__.py
+++ b/cryojax/experimental/__init__.py
@@ -1,3 +1,4 @@
+from ..simulator._api_utils import make_linear_operator as make_linear_operator
 from ..simulator._multislice import (
     AbstractMultisliceIntegrator as AbstractMultisliceIntegrator,
     FFTMultisliceIntegrator as FFTMultisliceIntegrator,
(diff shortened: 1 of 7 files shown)


DSilva27 (commented):
I think this is not ready to be in cryoJAX. I don't really see the immediate utility of something like this. The only thing I can think about is the tranpose being used for backprojection, but on a quick test it didn't work as I would expect.

```python

image_model = cxs.make_image_model(
    voxel_volume,
    ...
)

operator, vector = cxs.make_linear_operator(
    simulate_fn=lambda x: x.simulate(),
    args=image_model,
    where_vector=lambda x: x.volume_parametrization.fourier_voxel_grid
)

fourier_backproj = operator.transpose().mv(image).volume_parametrization.fourier_voxel_grid
```

I would expect `fourier_backproj` to be 3D Fourier voxel grid representing a single slice, and I should recover a flat voxel grid that looks similar to the image passed to the transpose operator. However, this is not really how it works.

At least a tutorial should be made showing how this can be useful; otherwise, it's a rejection from my part.

michael-0brien:
I agree, I think there are a lot of open questions about using this in practice. Let's move to cryojax.experimental and remove from the docs for now.

Issue #498, opened 2025-12-02
Utility function for rendering voxels from atoms


michael-0brien:
We should have a utility for converting atoms to voxels with something simple like

```python
import cryojax.simulator as cxs

path_to_pdb = ...
atom_volume = cxs.load_tabulated_volume(path_to_pdb)
voxel_volume = cxs.atoms_to_voxels(atom_volume)
```

Can follow similar patterns to `load_tabulated_volume` and `make_image_model` utilities. In particular, will need an `cxs.AutoVolumeRenderFn` class similar to `cxs.AutoVolumeProjection`.

PR #505, opened 2025-12-06
Add `render_voxel_volume` and `AutoVolumeRenderFn`, utilities for rendering voxels from atoms

michael-0brien:
Addresses #481 and #498.


cryojax/simulator/_api_utils.py
```diff
@@ -441,6 +451,114 @@ def simulate_fn(volume, ...):
     return atom_volume
 
 
+@overload
```
DSilva27, line 474: I think this is really useful. Maybe not for this PR, but we should consider making it something similar to `make_image_model`. For example, we should create an object called `volume_config` similar to `image_config`. That way we can define the render function as

```python
def render_voxel_volume(
    atom_volume,
    volume_config,
    render_fn=None
): ...

Then this function would check which type the atom potential is and make the appropriate render_fn. Similar to how `volume_integrators` are treated in `make_image_model`.

DSilva27 (approved):

cryojax/simulator/_api_utils.py
DSilva27: This is partly solved by the `AutoVolumeRenderFn`, but then all the documentation for the arguments other than `shape` and `voxel_size` gets buried. Separating the voxel grid configuration from the details on how to compute the intensities should make things simpler. It would also make the API similar for both images and volumes.

michael-0brien: Yeah the arguments getting buried is an unfortunate part of things. I’m a little unsure how the volume_config would work but point taken!

michael-0brien: The issue is that most configuration options depend on how the intensities are computed. I think we should add to the `AutoVolumeRenderFn` docstring before merging and describe what exact render_fn a given volume resolves to and link to each respective documentation for using more advanced options.

PR #521, opened 2025-12-17
cryoJAX 0.5.3 release

michael-0brien:


cryojax/simulator/_api_utils.py
```diff
@@ -92,6 +93,7 @@ def make_image_model(  # pyright: ignore[reportOverlappingOverload]
     transform: AbstractImageTransform | None = None,
     normalizes_signal: bool = False,
     signal_region: Bool[NDArrayLike, "_ _"] | None = None,
+    signal_centering: Literal["bg", "mean"] = "mean",
```
DSilva27, line 96: Centering is confusing; maybe something like ``normalization_mode`` would be better. If I see "centering" I'll immediately think that the signal will be centered in the box.

cryojax/simulator/_api_utils.py
```diff
@@ -92,6 +93,7 @@ def make_image_model(  # pyright: ignore[reportOverlappingOverload]
     transform: AbstractImageTransform | None = None,
     normalizes_signal: bool = False,
```
DSilva27, line 94: I wonder if we should clump all of this into a dictionary, something like `normalization_config = {region: ...., centering: ...} = None`. I find it kind of weird that if `normalizes_signal` is False and I set something like `signal_centering` to a non-default value, there is no effect. Alternatively, we can just raise a warning when this is the case.

michael-0brien: Yeah I agree. I considered messing around with this more, like having a "normalize_mode" argument instead of two arguments, where one only matters if the other is True. The issue was it made the API more confusing to use.

For example, I pushed a commit that had `normalize_mode='bg_subtract'`, `normalize_mode='none'`, and something else for the usual case. But I found what exactly to call the string very confusing, especially since the options refer only to the constant offset.

Maybe it would be good to pass a dict. Since this is a non-API breaking / minimal change, we can probably push ahead for now. It's unclear to me what the best way to go is---let's keep this in mind for the future.

PR #523, opened 2025-12-19
Bug fix in `GaussianWhiteNoiseModel` and `GaussianColoredNoiseModel`: noise is always normalized to variance 1

michael-0brien:
Patch fix for bug I found in the noise model classes. I think when `normalizes_signal = True`, this may have accidentally been normalizing the noise, as well as the signal. However, this should have been very obvious from simulation if this were happening, so we should look into this further.


michael-0brien:
Upon further thought, this should not have impacted code that controls SNR through `signal_scale_factor`. Thankfully this is the more common pattern we use

PR #524, opened 2025-12-23
implement parameter file and stack dataset for cryosparc data

DSilva27:


DSilva27:
While trying to implement the conversion from cryosparc to relion, I noticed that the poses in the new relion dataset are a bit off.  @michael-0brien could you make sure the pose from cryosparc is being loaded properly? I had to make a small change in  the `make_pytrees` function for cryosparc. Basically, now I always invert the pose because inverting for the  Axis-Angle pose is not as simple as negating. 

It's a bit difficult to track in the code the conversion for the pose inversion, so I'm not sure if I'm doing it correctly.

DSilva27:
It might just be numerical errors in the rotation code, but here are the differences between the original cryosparc dataset, the relion dataset obtained through pyem, and the relion dataset converted through the function I implemented:

Notice that even for the cs vs relion (pyem) the quaternion is a bit off, although it is much worse for relion (new)

image

DSilva27:
On a second note, this is just an error of 0.02-0.04 degrees (checked the angle between rotation matrices)

michael-0brien:
> It might just be numerical errors in the rotation code, but here are the differences between the original cryosparc dataset, the relion dataset obtained through pyem, and the relion dataset converted through the function I implemented:
> 
> Notice that even for the cs vs relion (pyem) the quaternion is a bit off, although it is much worse for relion (new)

So the conversion is done as `euler_pose = cxs.EulerAnglePose.from_rotation(aa_pose.rotation)`. Under the hood this uses a quaternion representation as intermediate and calls `SO3.log` to go from aa to quat. Conversely, euler to aa would use `SO3.exp`. This is all copied directly from `jaxlie` and is one of the parts of the code that I don’t fully understand. I think we’ll have to investigate these functions. I would also just check if single vs double precision impacts things.

DSilva27:
aah good point, I'll check double precision

michael-0brien:
I’ll also note that I know these functions in `SO3` do have some tricks in them to maintain differentiability, so I suspect that implementations won’t look identical to pyem. I think they still should be accurate though?

michael-0brien:
> Basically, now I always invert the pose because inverting for the Axis-Angle pose is not as simple as negating.

You could do 

```python
cpu = jax.devices(backend=“cpu”)[0]
with jax.default_device(cpu):
    inverse_pose = jax.device_get(aa_pose.to_inverse_rotation())
```

This converts to a quat, inverts, converts back to aa, then converts everything back to numpy copy-free.

DSilva27:
Confirmed that it's a numerical precision issue. Still, less than 0.1 degree error is pretty good.

michael-0brien:
> Confirmed that it's a numerical precision issue. Still, less than 0.1 degree error is pretty good.

We can use [`jax.enable_x64`](https://docs.jax.dev/en/latest/_autosummary/jax.enable_x64.html) to make sure it’s accurate in general, but I think this may have been added in a recent version of JAX

michael-0brien:
Closing for the [PR in `cryospax`](https://github.com/michael-0brien/cryospax/pull/22)!

PR #525, opened 2025-12-28
Tweaks to finalize `cryojax.ndimage` namespace

michael-0brien:
Biggest user-facing change is probably `cryojax.ndimage.ImageScaling` to `cryojax.ndimage.ScaleImage`. The latter is better as we add more transformations.


DSilva27 (approved):
Works well! Checked that the new scale image works as intended

michael-0brien:
> Works well! Checked that the new scale image works as intended

Thanks for the check!

PR #527, opened 2025-12-30
Image transforms for translation and rotation: `cryojax.ndimage.PhaseShiftFFT` and `cryojax.ndimage.RotateFFT`

DSilva27:


cryojax/ndimage/__init__.py
michael-0brien, line 1: Let's call these `PhaseShiftFFT` and `RotateFFT`. I know we currently have "fourier_image" as local variable names all over the code, but as we've started to have function / class names operating on FFTs, I'm slowly moving to real-space as `image` or fourier-space as `image_fft` (or just `fft` when appropriate). See the `rescale_fft` function in `cryojax.ndimage`.

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+            shifted_image = ifftn(shifted_image_fft).real
+            ```
+
+        """
+        self.translation_operator = jnp.exp(
+            -1.0j * (2 * jnp.pi * jnp.matmul(frequency_grid, offset))
+        )
+        self.is_rfft = is_rfft
+        if is_rfft:
+            self.shape = (frequency_grid.shape[0], (frequency_grid.shape[1] - 1) * 2)
```
michael-0brien: The issue here is that this is not always what the shape was in real-space and can be off by a pixel depending on parity. This is slightly problematic we don't know if the highest frequency in the x dim is self-conjugate or not. Instead you could just always assume `image` is square, get rid of `self.shape`, say `shape = (y_dim, y_dim)` in `__call__`, and put a warning in the docstring.

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+from ._base_transform import AbstractImageTransform
+
+
+class PhaseShiftFourierImage(AbstractImageTransform, strict=True):
+    """Apply a phase shift to an image in Fourier space.
+
+    This class implements a phase shift of an image in Fourier space.
+    The shift is specified in pixels or angstroms along each axis.
+
+    Attributes:
```
michael-0brien: Can get rid of Attributes in docstring. When adding this to the docs, you can just document `translation_operator` directly. `is_rfft` is redundant since it's in `__init__`.

I know this model is slightly ambiguous at what is an attribute; this approach is more user focused (and attributes are immutable anyway in JAX!)

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+    ):
+        """
+        **Arguments:**
+
+            - `offset`: The offset by which to shift the image, in pixels or angstroms.
+            - `frequency_grid`: The frequency grid in pixels or angstroms.
+            - `is_rfft`: Whether the frequency grid is in full or rfft format.
+                Right now only full format is supported for image rotation.
+
+        **Example:**
```
michael-0brien: Examples should go in class docstring, not `__init__`.

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+    def __init__(
+        self,
+        offset: Float[Array, "2"],
+        frequency_grid: Float[Array, "y_dim x_dim 2"],
+        is_rfft: bool,
+    ):
+        """
+        **Arguments:**
+
+            - `offset`: The offset by which to shift the image, in pixels or angstroms.
```
michael-0brien: Not sure if the extra indent here impacts the markdown. Can you give a shot at adding these transforms to the docs and render things locally to make sure things look right? For instructions, see CONTRIBUTING.md if you need.

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+
+    translation_operator: Complex[Array, "y_dim x_dim"]
+    is_rfft: bool
+    shape: tuple[int, int]
+    is_real_space: ClassVar[bool] = False
+
+    def __init__(
+        self,
+        offset: Float[Array, "2"],
```
michael-0brien: Type hints in the `__init__` should be `cryojax.jax_util.NDArrayLike` instead of `Array`, so that static type checkers don't complain if a numpy array is passed. Attributes should be an `Array` like you have, so then just run `jnp.asarray` in `__init__`.

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+    angle_degrees: float
+    frequency_grid: Float[Array, "y_dim x_dim 2"]
+    order: int
+    mode: str
+    cval: float | complex
+    is_real_space: ClassVar[bool] = False
+
+    def __init__(
+        self,
+        angle_degrees: float,
```
michael-0brien: A few comments on `__init__` and attributes.

1) `angle_degrees` and `pixel_size` should be attributes of type Float[Array, ""]. This allows them to work with JAX transformations, e.g. `RotateFourierImage` can be vectorized over them or calls under JIT won't recompile if they change. Additionally, `cryojax.jax_util.FloatLike` should be used for their `__init__` type hint and they should be cast with `jnp.asarray`. Additionally, `frequency_grid` can be type hinted with `NDArrayLike` and cast.
2) Keyword arguments should follow `FourierSliceExtraction` naming, i.e. `order` -> `interpolation_order`, `cval` -> `fill_value`, and `mode` -> `out_of_bounds_mode`. FWIW I don't mind getting rid of these arguments, or maybe just keeping `interpolation_order`. Whatever you find useful, we can always add them later.
3) You could get rid of `pixel_size` all together and force that `frequency_grid` is in nyquist units. Up to you.
4) Let's keep naming consistent between the angle here and the angle in `cryojax.ndimage.CylindricalCosineMask`. Unfortunately I've named it [`in_plane_rotation_angle`](https://github.com/michael-0brien/cryojax/blob/main/cryojax/ndimage/_transforms/_masks.py#L162C1-L162C50) there, which is a little long and not appropriate here. Maybe just `rotation_angle` and I'll change that at some point to the same?

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+    ):
+        """
+        **Arguments:**
+
+            - `angle_degrees`: The angle by which to rotate the image, in degrees.
+            - `frequency_grid`: The frequency grid in pixels or angstroms.
+            - `is_rfft`: Whether the frequency grid is in full or rfft format.
+                Right now only full format is supported for image rotation.
+            - `pixel_size`: The pixel size in angstroms or the unit of choice.
+            - `order`: The order of the spline interpolation. Only 0 and 1 are supported.
```
michael-0brien: A few more comments:

- "Spline" specifically refers to cubic interpolation
- On boolean docstrings:
    - `is_rfft`: If `True`, the frequency grid is X. Otherwise, Y.
- You can add reference to `map_coordinates` for the interpolation options:
    - `order`: ... See [`cryojax.ndimage.map_coordinates`][] for more information.

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+        else:
+            self.shape = (frequency_grid.shape[0], frequency_grid.shape[1])
+
+    @override
+    def __call__(
+        self, image: Complex[Array, "y_dim x_dim"]
+    ) -> Complex[Array, "y_dim x_dim"]:
+        """Apply the phase shift to the input image in Fourier space.
+
+        Args:
```
michael-0brien: Seems like some old docstrings in the `__call__` methods

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+    order: int = 1,
+    mode: str = "fill",
+    cval: float | complex = 0.0,
+) -> Complex[Array, "y_dim x_dim"]:
+    angle = jnp.deg2rad(angle_degrees)
+    c = jnp.cos(angle)
+    s = jnp.sin(angle)
+    rotation_matrix = jnp.array([[c, -s], [s, c]])
+
+    rotated_grid = frequency_grid @ rotation_matrix
```
michael-0brien: You are we sure we don't need to run `rotated_grid = jnp.reshape(frequency_grid.reshape((N, 2)) @ rotation_matrix, frequency_grid.shape[0:2])`, where N is `math.prod(frequency_grid.shape[0:2])`?

cryojax/ndimage/_transforms/_spatial_transform.py
```diff
@@ -0,0 +1,204 @@
...
+    N = rotated_grid.shape[1]
+    logical_frequency_coordinates = (rotated_grid * N) + N // 2
+    k_y, k_x = jnp.transpose(logical_frequency_coordinates, axes=[2, 0, 1])
+
+    image_fft = jnp.fft.fftshift(fourier_image)
```
michael-0brien: Your fftshifts look buggy to me. It seems you only shift the image, not the grid, and you don't shift back to the original convention.

- If your FFT is the full fourier plane, before `map_coordinates` you need `jnp.fft.fftshift(frequency_grid, axes=(0, 1))` as well as your `jnp.fft.fftshift(image_fft)`. This puts zero frequency in the center so that out of bounds indexing in the interpolation is correct. Then, you just need to return `jnp.fft.ifftshift(rotated_fft)`.
- If your FFT is the half plane (i.e. rfft), instead just run `jnp.fft.fftshift(image_fft, axes=(0,))` and `jnp.fft.fftshift(frequency_grid, axes=(0,))` before and `jnp.fft.ifftshift(rotated_fft, axes=(0,))` after.

cryojax/ndimage/_transforms/_spatial_transform.py
DSilva27: damm, copilot got me on this one. I didn't see it wrote spline, I only read interpolation

cryojax/ndimage/_transforms/_spatial_transform.py
DSilva27: I'll leave it as `in_plane_rotation_angle` and if you end up changing it we can change it here as well.

cryojax/ndimage/_transforms/_spatial_transform.py
DSilva27: I don't think it matters whether you have a shape like `(B, B, 2)` or `(B*B, 2)`. The operation only happens in the last axis; the shape of the other axis should be irrelevant.

cryojax/ndimage/_transforms/_spatial_transform.py
michael-0brien: Way ahead of you, I already changed the other one!

cryojax/ndimage/_transforms/_spatial_transform.py
DSilva27: I honestly find the `map_coordinates` function to be really unintuitive. I basically brute forced this one until I found the combination that worked.

michael-0brien: I relate to this. Also see the new `cryojax.rotations.SO2` class. may be nice to put this to use under the hood, though it’s basically just an API for the exact code you’ve written.

michael-0brien:
Looks pretty much good to go! Two things: 1) you can remove the `is_rfft` from the `__init__` doc and 2) you just need an `ifftshift`, not an `fftshift` after interpolation (could this have been why the real-space version wasn’t working?)

michael-0brien (approved):
I think things look fine to merge but the fact that `ifftshift` breaks things makes me suspect there could still be a bug. If you’d rather just merge and save this for another day ok by me.

michael-0brien:
@DSilva27 I was able to debug the rfft case. I added to docs and got tests running, so I think we are ready to merge.

PR #530, opened 2026-01-06
`CRYOJAX_ENABLE_CHECKS` environmental variable controls `equinox.error_if` run-time checks

### LINKED ISSUE #529: `equinox.error_if` calls can hurt computational efficiency
We call `equinox.error_if` in various `__init__`s, like in the pose, CTF, and volumes in order to check that arrays are bounded correctly. This wraps `jax.lax.cond`, which is inefficient on GPU and may be hurting some users performance if they are initializing objects after JIT boundaries. There are two improvements to computational efficiency I'd suggest:

- When equinox==0.13.3 is released make [EQX_ON_ERROR=off](https://github.com/patrick-kidger/equinox/pull/1134) the default for cryoJAX. Runtime error checking is only performed if the user wants.
- We currently have calls on arrays passed to atom volumes that make sure array values are non-negative (gaussian amplitudes, B-factors). This scales as $O(N_{atoms})$ (even if EQX_ON_ERROR=off), which is not a good pattern. I propose removing these checks all together.

In the meantime, users should avoid instantiating certain cryoJAX objects across JIT boundaries for more performant code, or they should install the equinox dev branch (at the time of writing this) to access the EQX_ON_ERROR=off option.

michael-0brien:
See #529. Need to wait until equinox==0.13.3 release to fully address this.


michael-0brien:
I've reworked the solution proposed in #529 to be more featureful than I initially thought was appropriate. I created an environmental variable `CRYOJAX_ENABLE_CHECKS`, which is `False` by default, that controls whether or not `equinox.error_if` is called.

I also did a bit of a reworking of the checks themselves; they are now performed at call time, not `__init__` time. The reason is that checks can be performed even after the user mutates states with `equinox.tree_at` or `equinox.apply_updates`.

PR #555, opened 2026-01-30
Directly compute correct mode ordering in `FFTAtomProjection` non-uniform FFT 

michael-0brien:
Eliminates the use of `fftshift` under the hood by going into the FINUFFT advanced options under the hood. Throw error messages now if the user tries to change to the wrong mode ordering under the hood.


michael-0brien:
Undoing this PR so removing from release notes

PR #567, opened 2026-03-20
Fix incorrect `FFTAtomProjection` reverse-mode gradients

### LINKED ISSUE #566: Bug in `FFTAtomProjection` reverse-mode autodiff
The `modeord` argument must be set to `False` for the backward (or `nufft2`) pass.

michael-0brien:
Backwards pass needs `modeord = False`. Recently I changed `modeord = True` for the simulator to avoid the extra `fftshift`, and incorrectly I also set it to `True` for the backwards pass. This also fixes an issue where gradients crash when setting `upsample_factor = ...` in this object, as the underlying call to `equinox.nn.Pool` was needed a compiler condition to be met in order to be differentiable.


michael-0brien:
Redoing in #569

PR #569, opened 2026-03-31
Fix incorrect `FFTAtomProjection` reverse-mode gradients

michael-0brien:
Redoing #567. See #566.


michael-0brien:
Tests passing, docs not building

michael-0brien:
Tests passing but docs not building. Will fix on version release

PR #573, opened 2026-05-07
cryoJAX 0.5.6 release

michael-0brien:
Biggest change is simplified treatment of the rotation convention. Also gradient bug fix to the `FFTAtomProjection` class. For merge, need to debug why docs are failing to build in CI.


DSilva27 (approved):
Looks good!

PR #574, opened 2026-05-19
Update `docs` dependencies to fix docs build failing

DSilva27:


DSilva27:
Another option (instead of forcing an old version of pygments) is to upgrade the version we are using for everything else in the docs dependencies. Specifically pymdown-extensions as it is currently at 10.14.3 while the latest version is 10.21.3.

DSilva27:
Decided to copy Equinox new pyproject.toml which also fixes the issue

michael-0brien (approved):
Nice. Thanks for figuring this out,

PR #575, opened 2026-05-19
More efficient approach to computing images en voxel grids in GaussianMixtureVolume: use `jnp.einsum` instead of `jnp.tranpose` followed by `jnp.sum`

DSilva27:
Here is a better approach to compute images and voxel grids through a `GaussianMixtureVolume`. Before this, we would evaluate the Gaussians along the x-, y-, and z-axes (only for voxel grids), create copies via transpose operations, and, in some cases, create new axes. Now we simply use `jnp.einsum`, which should be more efficient and avoids creating spurious copies.

In addition, this PR fixes a bug where coordinate grids generated by `AbstractImageConfig` had incorrect scaling in some pixels when requesting a physical grid. Frequency grids were unaffected.

- Improved performance of evaluation of images and voxel grids using `GaussianMixtureVolume`.
- Bug fixed when computing physical coordinate grids with `AbstractImageConfig`.

### DIFF
--- a/cryojax/simulator/_image_config.py
+++ b/cryojax/simulator/_image_config.py
@@ -212,7 +212,9 @@ class AbstractImageConfig(eqx.Module, strict=True):
 
         if physical:
             pixel_size = error_if_not_positive(self.pixel_size)
-            coordinate_grid = _safe_multiply_by_constant(coordinate_grid, pixel_size)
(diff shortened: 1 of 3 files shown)


michael-0brien (approved):
Looks great!

Issue #578, opened 2026-06-06
Migrate `jax-finufft` calls to `nufftax`?


michael-0brien:
`nufftax` is a pure-JAX implementation of the `finufft` algorithm: https://github.com/GragasLab/nufftax. This is great as the underlying `jax.ffi` calls via `jax-finufft` has issues and limitations like:

- difficult installation
- no native multi-GPU support
- JIT cache misses
- limitations in JAX compiler optimizations
- ...

So this is great! We should adopt this as a dependency, and also perhaps even help keep it maintained.

michael-0brien:
See #580 and #583. Let's keep an eye on this, it's a great package, but there's a few reasons why sticking with FINUFFT is best for now.

michael-0brien:
#584

Issue #285, opened 2024-12-06
Move `HighEnergyScatteringTheory` and `MultisliceScatteringTheory` out of `cryojax.experimental`


mjo22:
For this, we will need to run some tests. Also, need to somehow document the derivation of how the amplitude contrast ratio enters into the `WaveTransferFunction`

mjo22:
Progress in PR #356

mjo22:
Progress in #383

mjo22:
We should be ready to move the following objects from `cryojax.experimental`: `AbstractWaveScatteringTheory`, `HighEnergyScatteringTheory`, `WaveTransferTheory`, and `EwaldSphereExtraction`.

To do this, it would be good to add tests and documentation.

mjo22:
@geoffwoollard if you also want to take the lead on other methods in `cryojax.experimental` too, a road to the finish line likely looks like the following:

1) Resolve #416 (amplitude contrast for multislice). Depending on what we conclude, we can either do something simple and short-term or something more accurate and wait to move out of `cryojax.experimental`.
2) Resolve #417 (tests for scattering theories). See discussion there, probably just small adjustments to tolerances needed.
3) Correctly implement amplitude contrast ratio for the `HighEnergyApproximation` + `EwaldSphereExtraction` (i.e. thick phase grating approximation).
4) Implement test in `test_direct_integrators.py` for `EwaldSphereExtraction`. Maybe this could say for increasing wavelength at varying CTF parameters, does the MSE between `EwaldSphereExtraction` vs `FourierSliceExtraction` converge?

PR #579, opened 2026-06-06
cryoJAX 0.6.0 release

michael-0brien:
Roadmap to 0.6.0:

- [x] Prune deprecated code and change deprecation warnings to errors
- [x] Move `EwaldSphereExtraction` out of `cryojax.experimental` and add tests (#285)
- [x] Migrate `jax-finufft` to `nufftax` (#578)


michael-0brien:
Rather than migrating `jax-finufft` to `nufftax`, we add the `backend = ...` argument to the relevant projection methods. `jax-finufft` is of course highly optimized and can be more advantageous depending on the problem.

michael-0brien:
All relevant PRs for the release are now ready for review. Maybe there are smaller issues people want to tackle as well, but after these come through we should be good to go.

michael-0brien:
Can review and do the release after merging #594 and #598

michael-0brien:
If testing out the code, particularly if using the `GaussianMixtureVolume` class:

- Try setting the new `n_spread` variable in `GaussianMixtureProjection` for a more efficient algorithm for gaussian spreading. You can use the helper `cxs.suggest_n_spread` to figure out what you should set it to. This will particularly be advantageous for large numbers of atoms, and for pixel sizes much larger than gaussian widths.
- After doing this, if you’re running on GPU set the envvar `CRYOJAX_ENABLE_PALLAS=1`. This should result in performance upgrades in memory-constrained code, or solid upgrades for `jax.grad` calls.

tests/test_ndimage.py
```diff
@@ -18,25 +20,122 @@ def test_downsample_preserves_sum(shape, downsample_factor):
...
+    coarse_gaussian = jnp.exp(-jnp.sum(coarse_grid**2, axis=-1) / (2 * sigma**2))
+
+    np.testing.assert_allclose(downsampled, coarse_gaussian, atol=2e-2)
+
+
 #
 # FFT
 #
 @pytest.mark.parametrize("shape", [(10, 10), (10, 10, 10), (11, 11), (11, 11, 11)])
 def test_fft_agrees_with_jax_numpy(shape):
```
DSilva27: This test should be removed. It used to compare `jnp.fft` with `cryojax.ndimage.fft`, but now it just compares `jnp.fft`against itself.

DSilva27 (approved):

tests/test_ndimage.py
michael-0brien: good catch! fixed

michael-0brien:
Found minor bug in new RFFT fourier slice code that only showed up for certain box dims; $(-1)^k$ phase pattern was applied to the wrong mode ordering. Fixed

PR #580, opened 2026-06-06
Replace `jax-finufft` calls with `nufftax`

michael-0brien:
Would be good to do benchmarks before merging. Need to address https://github.com/GragasLab/nufftax/issues/8.

Additionally, we can give another shot at setting `modeord=1` to avoid the extra FFT shifts.


michael-0brien:
Let's also change `FFTAtom*` to `IndependentAtom*` to match the volume name and futureproof what the underlying implementation may be.

michael-0brien:
Was getting several errors out of the box on blackwell hardware, despite basic correctness tests working. Given finufft is highly optimized and nufftax is very new, I think it's best to wait. In the meantime, I'll setup a PR that enables us to implement our own projection method for direct spreading onto the plane.

PR #583, opened 2026-06-07
Rename `FFTAtom*` pattern to `IndependentAtom*` and setup API / backend for fast real-space atom rendering

michael-0brien:
See https://github.com/GragasLab/nufftax/issues/10. We could adapt `nufftax` or PR more general capability for spreading atom kernels onto the plane. Made the relevant backend changes in this branch so we can implement this cleanly.


michael-0brien:
Tried testing `nufftax` at scale and got a few errors, including OOM errors and nans. Don't have the time at the moment to sort out in detail why, but I do think the value of how optimized and stable FINUFFT is outweighs the pure JAX implementation.

Instead, I prepared an API for us to implement our own direct real-space atoms to pixels/voxels.

michael-0brien:
See here: https://github.com/GragasLab/nufftax/pull/11; I would propose using nufftax for the real-space method in a separate PR, and perhaps adding a nufftax backend for the current method. We would need to get a minor PRs merged, e.g. I think for now it is best to opt into their optimized GPU kernel backend for stability reasons, but this would be a great way to go!

PR #584, opened 2026-06-08
Adopt `nufftax` for `IndependentAtom*` rendering/projections, including new real-space method

### LINKED ISSUE #578: Migrate `jax-finufft` calls to `nufftax`?

michael-0brien:
We can adopt `nufftax` in the near future to support usage of `IndependentAtomVolume` without need to install `jax-finufft`. Since FINUFFT is highly optimized, we keep this as an option via `IndependentAtomProjection(..., backend="jax-finufft")`.

Additionally, https://github.com/GragasLab/nufftax/pull/11 adds the ability to spread custom kernels onto the plane, and #583 sets up the API / backend for a real-space method. If real-space kernels are passed, we can use `nufftax.core.spread_*d` functions to directly spread gaussians onto the plane.

### DIFF
--- a/cryojax/ndimage/__init__.py
+++ b/cryojax/ndimage/__init__.py
@@ -35,6 +35,7 @@ from ._fourier_statistics import (
 from ._fourier_utils import (
     convert_fftn_to_rfftn as convert_fftn_to_rfftn,
     enforce_rfftn_self_conjugates as enforce_rfftn_self_conjugates,
+    query_efficient_grid_size as query_efficient_grid_size,
 )
 from ._map_coordinates import (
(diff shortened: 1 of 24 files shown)


michael-0brien:
Have a solid draft finished. Will open this for review when PRs are merged in nufftax and a new version is released.

PR #586, opened 2026-06-12
Update branch

michael-0brien:
I found a while ago that my reconstructions in RELION weren't working. This may be why? Box size of odd image dim has zero in the center [-L/2, L/2], but we had zero where FFTs place their zero frequency. In reality, x = 0 is actually between two pixels.

This explained why it was a big pain to get finufft to match our conventions how we expect as well, so this PR includes some improvements and optimizations on that front now that conventions match. This is a somewhat significant PR, so let's take time to test things and merge.


michael-0brien:
Returned to previous convention because we do match RELION's stated convention (https://relion.readthedocs.io/en/release-3.1/Reference/Conventions.html).

I was able to make projection methods match across conventions, so lumping this PR into #584, which includes various other code optimizations.

PR #587, opened 2026-06-12
Reduce XLA scatters for GPU efficiency; fix bug in real-space coordinate generation

michael-0brien:
While creating more efficient implementations of some commonly used functions of ours, found that a bug in real-space coordinate generation was merged into v0.5.6 where the pixel size was not being multiplied in. This will have to be flagged and we'll have to prepare a new release soon.


michael-0brien:
We can merge in other branches first though! This PR is branched off of real-space-rendering.

michael-0brien:
Flagged that #587 introduced bug in fourier slice; fixed here

michael-0brien:
Failing CI fixed upstream

PR #593, opened 2026-07-01
Scattering theory features: move `EwaldSphereExtraction` out of `cryojax.experimental`

### LINKED ISSUE #285: Move `HighEnergyScatteringTheory` and `MultisliceScatteringTheory` out of `cryojax.experimental`

michael-0brien:
Adds the relevant tests, and also renames `StrongPhaseScatteringTheory` -> `RytovScatteringTheory` for clarity. I think in practice this was still an experimental module so the name change is ok and will be considered stable in v0.6.0.

I think after this PR will be ok to close #285


michael-0brien:
Note that #590 should be merged first

DSilva27 (commented):
Looks good. Most changes are related to the removal of Deprecations, which I already checked. The other changes are very minor: Ewald is out of experimental, unitest for it, and some objects are renamed.

PR #594, opened 2026-07-04
`GaussianMixtureVolume` gets short-ranged spreading backend

michael-0brien:
This PR adds an adapted version of the nufftax spreading code the repository, and gives `GaussianMixtureVolume` the ability to call it through a parameter called `n_spread`.

Also includes general API and backend cleanups:

- ~~All mentions of `eps` from other classes are now changed to `n_spread` for spreading kernels as this is more clear and general~~ (walked this back in #598)  
- `batch_options` is now just `n_batches` for `GaussianMixtureRenderFn`, consistent with the projection API

Note that #593 and #590 should be merged first (I branched off of these PRs, so commits + changes look little confusing in the PR)


michael-0brien:
So I've added additional major upgrades via `jax.experimental.pallas` in this PR. Custom GPU kernels greatly reduce memory footprint and make a very efficient backwards pass. The forward pass via pallas is comparable on H100 but usually slower, so we allow independent forward and backward tuning.

Turn on this behavior with `CRYOJAX_ENABLE_PALLAS=1` envvar (or locally `GaussianMixtureVolume(..., enable_pallas=...)`)

cryojax/ndimage/_spreading/spread.py
```diff
@@ -0,0 +1,395 @@
...
+
+def _support_indices_and_offsets(
+    coord: Float[Array, " M"], n: int, n_spread: int
+) -> tuple[Array, Array]:
+    """Per-axis spread indices (wrapped to `[0, n)`) and kernel offsets `z`."""
+    i0 = jnp.ceil(coord - n_spread / 2.0).astype(jnp.int32)
+    offsets = jnp.arange(n_spread)
+    indices = i0[:, None] + offsets[None, :]  # (M, n_spread)
+    z = indices.astype(coord.dtype) - coord[:, None]
+    return indices % n, z
```
DSilva27, line 151: Not sure if I understand the code correctly, but wouldn't `indices % n` create periodic boundary conditions? Is that the desired behavior when a Gaussian is spread to the edge of the box? This should rarely be a problem if the box size is set appropriately for the particle, but I'm still worried this would be undesired behavior.

I know the issue is that we need a fixed shape to be able to jit, so maybe instead we could add something like `valid = Bool(M, n_spread)`, and then we simply multiply by `valid` before doing the segment sum (i.e., invalid entries don't contribute towards the projection).

michael-0brien: I would say this is the behavior we want; imagine our gaussian density is very small at the edges but non-zero—if we truncate we will pick up small numerical error in our FFTs (coefficients are periodic, so there will be a small discontinuity).

If people are in an aphysical regime, they should increase their padded_shape

PR #598, opened 2026-07-07
Finalize `IndependentAtomVolume` -> `GaussianFourierVolume` (removes real-space capability)

michael-0brien:
With a more efficient GaussianMixtureVolume, we should have each class specializing in real vs fourier (and make this clear in the name). `IndependentAtomVolume`'s real-space capability was just a worse version of `GaussianMixtureVolume` after #594.


michael-0brien:
Marking as draft until #594 (upstream) is merged. In theory though this is ready for review

PR #602, opened 2026-07-08
Remove `RealVoxelCloudVolume` and `RealVoxelProjection`

michael-0brien:
This was just maintenance overhead, and it doesn't fulfill a particular niche where it is better than fourier slicing. To my knowledge it is not being used by anyone.


DSilva27 (approved):
Looks good. These were not needed anyway.

PR #603, opened 2026-07-08
Major usability upgrades: instantiate classes with batch dims, do not transfer to device during initialization

michael-0brien:
This was making downstream code awkward to use. Often you don’t want to transfer to device until you hit a JIT region.


michael-0brien:
@DSilva27 after this in cryospax, we will no longer need tree_at and jax.default_device calls

michael-0brien:
Marking as draft until upstream branches are merged

